PR #2, opened 2015-06-09
[HWKMETRICS-106] Upgrade embedded Cassandra to 2.1.5

stefannegrea:


pilhuhn:
While I do not see any issues with this, I think that should be communicated on dev-list, as other parties now also use Cassandra (or will use it).

PR #7, opened 2015-06-23
Adjust package names to comply with path structure.

pilhuhn:


pilhuhn:
Respective change in hawkular: master  http://fpaste.org/235757/14350684/

PR #9, opened 2015-06-24
Allow to override the backend sys-prop from the environment.

pilhuhn:
This is/can be used to prevent/enable starting the embedded C* database from an env variable.
This way, when using external C* nodes on Docker one can start the docker image with an env of
`-e HAWKULAR_BACKEND=""` to prevent the startup of embedded C*.


ppalaga:
lgtm

PR #11, opened 2015-08-06
Fix imports order to match the new Checstyle rule

ppalaga:


ppalaga:
@pilhuhn could you please review or eventually delegate to another more suitable reviewer?

pom.xml
```diff
@@ -130,6 +130,14 @@
         <version>${version.org.apache.cassandra}</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.wildfly.bom</groupId>
+        <artifactId>jboss-javaee-7.0-wildfly</artifactId>
+        <version>${version.org.wildfly}</version>
```
pilhuhn: Would it make sense to limit that to the module that defines javax.ejb.* as it looks like no other EE feature is used

ppalaga: Should any part of commons run in a container distinct from WF with `${version.org.wildfly}`? Otherwise, I would not want to manage javax.ejb.* separately.

pilhuhn: On 6 Aug 2015, at 11:09, Peter Palaga wrote:

>> @@ -130,6 +130,14 @@
>>       <version>${version.org.apache.cassandra}</version>
>>     </dependency>
>>
>> +      <dependency>
>> +        <groupId>org.wildfly.bom</groupId>
>> +        <artifactId>jboss-javaee-7.0-wildfly</artifactId>
>> +        <version>${version.org.wildfly}</version>
>
> Should any part of commons run in a container distinct from WF with 
> `${version.org.wildfly}`? Otherwise, I would not want to manage 
> javax.ejb.* separately.

In fact we only need

import javax.ejb.Singleton;
import javax.ejb.Startup;

which is EJB 3(.1) and not WF-specific at all.

PR #14, opened 2015-08-17
Create CassandraDaemon as runManaged

lucasponce:
In C* 2.2.0 a call to CassandraDaemon.deactivate() will call to a System.exit(0) if runManaged flag was not activated.
This may lead in troubles if when this call is invoked inside an app server.


pilhuhn:
Good catch

jshaughn:
I will take you word for it :-)

PR #16, opened 2015-09-23
HAWKULAR-628 Let embedded Cassandra service manage its own logging

tsegismont:
HAWKULAR-628 Let embedded Cassandra service manage its own logging

Excluded logging subsystem and provided a logback config file.

Cassandra logs now go to "${jboss.server.log.dir}/embedded-cassandra.log". Log files follow a time base rolling policy (similar to Wildfly server log file).

It's still possible to configure Cassandra log level via the "hawkular.log.cassandra" system property.

Also:
- removed log4j config (Cassandra uses logback now)
- removed unused POM config
- excluded log4j and junit deps


ppalaga:
@tsegismont I checked that your PR is doing what you promise, which is big improvement against what we have now. Thanks for that. 

I have sent a little enhancement against your branch https://github.com/tsegismont/hawkular-commons/pull/1 It makes the emebedded cassandra to send ERR and WARN messages to stdout so that they appear in the WF console. The messages do not look nice, because jboss logging prefixes them with `date INFO  [stdout] (thread)` but it is IMO better to see them like this rather then to oversee them in the separate log file. I am ready to merge the present PR either with or without this enhancement. Feel free to decide if you find it good enough to merge to your branch.

tsegismont:
As you wish.

PR #22, opened 2015-11-12
[HWKINVENT-158] Introduce a generic `/status` REST handler sharable by all hawkular components.

metlos:


ppalaga:
@metlos the first round of review finished.

Is there a wide consensus that commons is the right place for this? Why not the new nest repo?

metlos:
> @metlos the first round of review finished.
>
> Is there a wide consensus that commons is the right place for this? Why not the new nest repo?

Not actually sure. But since this is supposed to be consumed by components, I am a little bit nervous about putting it in nest to avoid circular deps we've seen. That said, I'm open to any suggestions..

PR #25, opened 2015-11-24
Add ManifestUtilTest

ppalaga:


ppalaga:
@metlos  could you please review and eventually merge?

PR #26, opened 2015-11-24
Package cassandra driver as a WildFly feature pack

ppalaga:


ppalaga:
This is not ready at all. Putting here to share with others, esp. @jpkrohling .

ppalaga:
Replaced by a module in Nest module of Commons

PR #30, opened 2015-12-10
Introduce our own Guava WildFly module containing Guava 16.0.1

ppalaga:


ppalaga:
@jpkrohling could you please review an eventually merge?

jpkrohling:
I'll merge, but I think we should consider this one "experimental". I have no idea what could be the side effects of this :)

PR #35, opened 2015-12-18
HAWKULAR-891 Missing dependency on C* driver module

lucasponce:
Fix for a blocker issue on alerts scenario.


pilhuhn:
Is `javax.api` the module name from jboss modules for all the `javax.*` packages?

lucasponce:
Probably, I follow the suggestion of Tomaz Cerar on this comment:
https://developer.jboss.org/thread/230525?tstart=0
As it is the same issue and I checked that is solved with this.

PR #36, opened 2015-12-21
HAWKULAR-732

jsanda:
Long overdue bus refactoring that makes use of JMS 2.0 APIs as well as CDI. There is still some work to due such as adding/updating tests and updating logging, but I would like to start the review on this since I have had the branch open for a while now. I should also note that I have branches in my forks of hawkular-command-gateway and hawkular-inventory to refactor them to use the new bus API, which is another reason this branch has been lingering.


ppalaga:
@jsanda is there a reason to keep so many commits here? It would be much easier to review with less, ideally a single commit.

ppalaga:
@jsanda I think it would be practical to have at least one working itest here before proceeding with cmdgw, inventory, etc.

jmazzitelli:
I talked to John about this before the holiday break. We are putting this PR on hiatus for a while. Some of this stuff I wasn't thrilled about (e.g. putting JMS dependencies in the BasicMessage hierarchy and other things around BasicMessage) but also there were issues with the agent itests failing with these changes and we weren't sure why the errors were occurring.

As I understand it, the reason why this whole push to refactor the bus API was started in the first place was to be able to use the JMS 2.0 feature of putting messages on the broker asynchronously. Some of the changes in here do not seem to be related to that. If that is the main crux of the problem to be solved, I suggest we rollback much of this and just concentrate on solving that one problem. I think it would possibly make the changes less drastic/severe and could be easier/faster to get done. And with less code changes, easier to peer review and less likely to cause breakages.

ppalaga:
> As I understand it, the reason why this whole push to refactor the bus API was started in the first place was to be able to use the JMS 2.0 feature of putting messages on the broker asynchronously. Some of the changes in here do not seem to be related to that (e.g. do we really need to introduce CDI in this PR? Do we really want to re-implement BasicMessage and JSON parsing 

I welcome this attempt to make the API (esp. BasicMessage/BasicMessageWithExtraData and JSON parsing) simpler and easier to use.

> and introduce a JMS dependency in BasicMessage

@jsanda indeed, a piece of JavaDoc for org.hawkular.bus.common.BasicMessage.getReplyTo() would be helpful. What is it good for?

jmazzitelli:
> I welcome this attempt to make the API (esp. BasicMessage/BasicMessageWithExtraData 
> and JSON parsing) simpler and easier to use.

Yes, my only question is do we NEED it versus is this just something nice to have? I saw that the BasicMessage/WithExtraData refactoring was nice. But all of this already works as-is (and the new refactoring currently is broken). Do we need to rewrite it for the sake of simplicity when there are other things that don't work or that need to get done (like what John is working on now - the OpenShift integration). So I ask if we should  concentrate on things that need to get done, rather than on what would be nice to get done.

The JMS 2.0 stuff could also be considered an optimization that we could put off for later (though I understand that putting messages on the broker asynchronously helps the performance a lot - so that is an important optimization we should eventually do - but I think that's a separate issue from adding CDI support and refactoring BasicMessage - neither of which is a pressing concern)

jsanda:
I do think that most of the changes should be done and not simply nice to haves; however, I wound up trying to bite off too much in this PR. I do want to revisit this, but it needs to be done more incrementally.

ppalaga:
@jsanda I have squashed your commits myself locally and I have a couple of comments that I'd like to put here. Could you please reduce the number of commits in this PR so that is easier for me to find the proper location for my comments?

jsanda:
I just squashed my commits including some uncommitted changes from prior to the holidays.

pilhuhn:
@jsanda @ppalaga @jmazzitelli What is the state of this (other than that it has a merge conflict)

ppalaga:
@pilhuhn I'd call it dead by starvation. I still (as I always did) find the goals pursued here for desirable.

jmazzitelli:
dead

jsanda:
RIP

PR #37, opened 2016-01-05
Remove the dev Maven profile that is not used anymore

ppalaga:


ppalaga:
@jmazzitelli this should fix the building of HK main with dev profile. Could you please review and eventually merge?

PR #40, opened 2016-01-14
Use version.org.jboss.arquillian from hawkular-parent

ppalaga:


ppalaga:

PR #41, opened 2016-01-14
RPC message listener -fix response connection closing

pavolloffay:
RPC message listener should not close listening connection instead it should close connection of a response.

PR sent to original repository: https://github.com/hawkular/hawkular-bus/pull/56


pavolloffay:
@jmazzitelli  Could you please review?

PR #42, opened 2016-01-18
HAWKULAR-945 - Do not throw exceptions on PostConstruct for Singleton

jpkrohling:


jpkrohling:
@ppalaga , similar to hawkular/hawkular-accounts#111 , if/when you have a few minutes :)

PR #43, opened 2016-02-03
Downgrade to wildfly 10.0.0.CR5 and wildfly.core 2.0.5.Final to make the itests 

ppalaga:


ppalaga:
@jpkrohling this to show what happens: just compare the result with and without the last commit

jpkrohling:
It seems that ActiveMQ allowed queues to be created on demand, and it seems that WF 10.0.0.Final removed this capability. So, adding a TestQueue to the `standalone-test.xml` seems to fix the tests.

On the following branch, I added a commit that synchronizes the standalone-test.xml with the latest WF 10.0.0.Final, so that it's easier to see the changes (although the change mentioned above could *not* be found on standalone.xml).

jpkrohling:
I guess I missed something: the link!
https://github.com/jpkrohling/hawkular-commons/tree/ppalaga-p35-with-changes

jpkrohling:
Confirmed that this is the case: https://github.com/wildfly/wildfly/pull/8620/files .

ppalaga:
@jpkrohling thanks for your investigations.

Yes, fixing the itests through adding the `TestQueue` to the `standalone-test.xml` would solve the problem here. But we need the ability to create the queues dynamically in Command Gateway. Therefore, I'd say that would be good to make the tests here to behave similarly as the Command Gateway. We need to either (a) enable the autocreation [1] or (b) create the queues ourselves over the Artemis management APIs [2]. 

[1] https://github.com/wildfly/wildfly/pull/8620/files
[2] https://activemq.apache.org/artemis/docs/1.0.0/management.html

jpkrohling:
I just changed the commit on the mentioned branch to set the boolean auto-create-jms-queues to true on the standalone-test.xml . But a similar change would be required for the actual standalone.xml .

jpkrohling:
By the way: it might be interesting to add `auto-delete-jms-queues` to true as well. But I don't know much about ActiveMQ, so, I can't say if that's really required for our setup.

ppalaga:
@jpkrohling ad5246c adds exactly what you proposed. The tests should pass now. Please re-review and eventually merge.

PR #45, opened 2016-02-10
Remove unneeded dependencies, sort dependencies alphabetically

ppalaga:


ppalaga:

hawkular-bus/hawkular-bus-common/pom.xml
```diff
@@ -37,6 +37,18 @@
   </properties>
 
   <dependencies>
+    <!-- To avoid javadoc errors (and thus causing mvn release plugin to fail) we need this provided dependency.
```
jpkrohling: Is this really needed? I've seen similar comments on the `hawkular/modules`, and the javadoc passed/failed regardless of the dependency.

hawkular-bus/pom.xml
```diff
@@ -79,28 +69,11 @@
...
       <dependency>
-        <groupId>org.wildfly.core</groupId>
-        <artifactId>wildfly-controller</artifactId>
-        <version>${version.org.wildfly.core}</version>
-      </dependency>
-
-      <dependency>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-dist</artifactId>
+        <artifactId>wildfly-jms-client-bom</artifactId>
```
jpkrohling: I'm sure this dependency is correct, but just to confirm: isn't this also included on the main wildfly-javaee bom?

jpkrohling:
Review done.

hawkular-bus/hawkular-bus-common/pom.xml
ppalaga: > Is this really needed? I've seen similar comments on the hawkular/modules, and the javadoc passed/failed regardless of the dependency.

No idea. This comes from you @jmazzitelli ? Can it eventually be removed?

hawkular-bus/pom.xml
ppalaga: Clicking through "usages" on http://mvnrepository.com/artifact/org.wildfly/wildfly-jms-client-bom/10.0.0.Final, it does not look like it was linked with wildfly-javaee bom. Neither I can see it anywhere when studying wildfly-jms-client-bom.

PR #47, opened 2016-02-17
HAWKULAR-1045 - Refactored BinaryData to use SequenceInputStream

jpkrohling:


jpkrohling:
@jmazzitelli , could you please review this one?

hawkular-bus/hawkular-bus-common/src/main/java/org/hawkular/bus/common/BinaryData.java
```diff
@@ -59,68 +59,52 @@ public void setOnCloseAction(Runnable action) {
...
 
-    public int read() throws IOException {
-        if (unreadInMemoryDataExists()) {
-            return (inMemoryData[inMemoryDataPointer++] & 0xff);
-        } else {
-            return streamData.read();
-        }
-    }
+    public void close() throws IOException {
+        backingStream.close();
```
jshaughn: do we need a try-finally here to ensure any close action gets called?

hawkular-bus/hawkular-bus-common/src/main/java/org/hawkular/bus/common/BinaryData.java
```diff
@@ -16,13 +16,14 @@
...
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.SequenceInputStream;
 
 /**
- * This is a "stream" that includes some in-memory data as well as another stream.  The in-memory data (if not null
- * or empty) will be read first. Once the in-memory data has been exhausted, any additional reads will read data
- * from the input stream.
+ * This is a stream that is backend either by the provided InputStream or by a SequenceInputStream, should the
```
jshaughn: "backend" -> "backed"

hawkular-bus/hawkular-bus-common/src/main/java/org/hawkular/bus/common/BinaryData.java
jmazzitelli: Yeah, that's a good catch. We should have been doing that before. We should add it now.

hawkular-bus/hawkular-bus-common/src/test/java/org/hawkular/bus/common/BinaryDataTest.java
```diff
@@ -16,10 +16,14 @@
  */
```
jmazzitelli: need to change the license header - add 2016 to pass license checker

hawkular-bus/hawkular-bus-common/src/main/java/org/hawkular/bus/common/BinaryData.java
jpkrohling: I don't think we should. We just delegate the calls, so, either the caller catches it, or the backing stream handles what is to be handled.

hawkular-bus/hawkular-bus-common/src/test/java/org/hawkular/bus/common/BinaryDataTest.java
jpkrohling: Done

hawkular-bus/hawkular-bus-common/src/main/java/org/hawkular/bus/common/BinaryData.java
jpkrohling: Fixed

hawkular-bus/hawkular-bus-common/src/main/java/org/hawkular/bus/common/BinaryData.java
jmazzitelli: He's talking about the close action. Once we delegate to the close(), we need to call that close action in order to clean up things. This was to clean up the temporary file that was produced when using the binary stream in ActiveMQ but it could be for anything - whether the close succeeded or failed, that needed to happen because there is no other way for the caller to later trigger that close action, hence why we should wrap this in a try..finally:

```java
try {
    backingStream.close();
} finally {
    // if we were asked to do something after close is done, do it now
    if (onCloseAction != null) {
        onCloseAction.run();
    }
}

jmazzitelli:
I guess you didn't build this locally :) license checker passes now, but checkstyle fails. see travis log

hawkular-bus/hawkular-bus-common/src/main/java/org/hawkular/bus/common/BinaryData.java
jpkrohling: Oh, right, sorry.

jpkrohling:
Now I did :D

PR #50, opened 2016-02-25
Make Checkstyle happy:  * Constructors in classes with package visibility should

ppalaga:


ppalaga:

Note that the second commit f1234db was necessary when I upgraded Checkstyle to v6.15 in the parent. But then, I found out that it causes more trouble in other components and I rather decided to stay with Checkstyle 6.8 in parent, so that I do not need to fix anything elsewhere. Nevertheless I am keeping f1234db here for the case that somebody will want to upgrade Checkstyle in the future.

jpkrohling:
LGTM

PR #51, opened 2016-02-25
Move subsystem-templates/nest.xml to the feature pack

ppalaga:


ppalaga:

As the next step I'd try to do the same thing in Accounts, unless you started already?

jpkrohling:
Is this going to Alpha 11 ?

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/configuration/standalone/hawkular-nest-template.xsl
```diff
@@ -0,0 +1,46 @@
...
+
+  <xsl:output method="xml" version="1.0" encoding="UTF-8" indent="yes" xalan:indent-amount="4" standalone="no" />
+  <xsl:strip-space elements="*" />
+
+  <!-- Add the Nest subsystem -->
+  <xsl:template match="/*[name(.)='server']/*[name(.)='extensions']">
+    <xsl:copy>
+      <xsl:apply-templates select="node()|@*"/>
+    </xsl:copy>
+    <system-properties>
```
jpkrohling: If there are properties defined in the "parent", those are then replaced, right?

hawkular-nest/hawkular-nest-feature-pack/pom.xml
```diff
@@ -96,6 +96,100 @@
...
+                <transformationSet>
+                  <dir>${project.build.directory}/feature-pack-resources-tmp</dir>
+                  <stylesheet>${basedir}/src/main/xsl/configuration/standalone/hawkular-nest-subsystems.xsl</stylesheet>
+                  <includes>
+                    <include>configuration/standalone/subsystems.xml</include>
+                  </includes>
+                  <outputDir>${project.build.directory}/feature-pack-resources</outputDir>
+                  <fileMappers>
+                    <fileMapper implementation="org.codehaus.plexus.components.io.filemappers.RegExpFileMapper">
+                      <pattern>^configuration/standalone/(.*).xml$</pattern>
```
jpkrohling: We only support standalone, right? Not `standalone-ha`, or `standalone-full-ha`. If that's still true, then this looks OK.

hawkular-nest/hawkular-nest-feature-pack/pom.xml
```diff
@@ -96,6 +96,100 @@
...
+                  <includes>configuration/standalone/template.xml,configuration/standalone/subsystems.xml</includes>
+                  <outputDirectory>${project.build.directory}/feature-pack-resources-tmp</outputDirectory>
+                </artifactItem>
+              </artifactItems>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <!-- Generate the standalone*.xml stubs from the upstream originals using XSL transformations -->
```
jpkrohling: This comment isn't completely true, I suppose. 

From what I could understand, each transformation set uses resources matching the pattern as input for the XSL, spitting out the resulting XML into feature-pack-resources. 

One transformation set is for subsystems.xml , the other is for template.xml.

jpkrohling:
Sometimes I wonder if it wouldn't be better to have a "descriptor" in a saner language somewhere outside of the `pom.xml`...

Anyway: review completed.

ppalaga:
> Is this going to Alpha 11 ?

Yes, I'd like it to be the case

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/configuration/standalone/hawkular-nest-template.xsl
ppalaga: > If there are properties defined in the "parent", those are then replaced, right?

Good question. Rather not replaced, but the result will contain <system-properties> twice. 

Do you think I should be more defensive in this specific XSL to cover also the case that the input already contains <system-properties> ?

hawkular-nest/hawkular-nest-feature-pack/pom.xml
ppalaga: Also a good question. Nobody asked for other than plain `standalone` yet. Once the requirement comes, we know how to add those other files.

hawkular-nest/hawkular-nest-feature-pack/pom.xml
ppalaga: > One transformation set is for subsystems.xml , the other is for template.xml.

Strictly speaking yes, but taken a little bit figuratively, both subsystems.xml  and template.xml are some kind of a stub used for building standalone*.xml files, are they not?

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/configuration/standalone/hawkular-nest-template.xsl
jpkrohling: No, as long as we can remember about it if/once we get this problem :)

hawkular-nest/hawkular-nest-feature-pack/pom.xml
jpkrohling: +1

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/configuration/standalone/hawkular-nest-template.xsl
ppalaga: Yes, I think the same :)

hawkular-nest/hawkular-nest-feature-pack/pom.xml
jpkrohling: They are, but I found it confusing at first and wondered if there was a `standalone.xml` there. If you want to keep the reference to XML files used to assemble a `standalone.xml` , then I'd reference `standalone/*.xml` instead.

jpkrohling:
I guess it should be OK to have it on Alpha 11. If there's a problem on this part, we'll see it very soon :)

hawkular-nest/hawkular-nest-feature-pack/pom.xml
ppalaga: OK, changed to `Enhance various *.xml files taken from our dependencies using XSL transformations`

jpkrohling: +1

jpkrohling:
If Travis is happy, I'm happy as well.

ppalaga:
@jpkrohling thanks!

PR #54, opened 2016-03-30
Upgrade to hawkular-parent 39 and stop delivering an own guava module

ppalaga:


ppalaga:
The motivation for replacing the old `ConfigEditor` with `CassandraYaml` is that we need the flexibility offered by `CassandraYaml` on other components (Accounts and Inventory at least) that do functional tests against embedded Cassandra.

ppalaga:
@jsanda and/or @stefannegrea could you please review?

jsanda:
As long as there is no regression, I am fine with the changes. What is the flexibility gained by ```CassandraYaml``` though? Just curious.

ppalaga:
>  What is the flexibility gained by CassandraYaml though?

It is simple to use from other components, e.g. here in Accounts: https://github.com/ppalaga/hawkular-accounts/commit/c3ed9c41d09d30cc00265e85cc19ee054f5c8e5a#diff-b0dd9bd68756e124a5fb77db828e32e0R250 Something like this was not possible with `ConfigEditor`. Accounts had to maintain its own `cassandra.yaml` plus it had to take care for the pre-existence of all the  cassandra cluster directories before starting the tests.

PR #56, opened 2016-04-14
Make embedded Cassandra use less memory

ppalaga:


ppalaga:
@jsanda could you please check how much this makes sense? I basically followed http://opensourceconnections.com/blog/2013/08/31/building-the-perfect-cassandra-test-environment/ which quite old (2013). Maybe you see other strategies to reduce the memory footprint of Embedded Cassandra?

ppalaga:
@jsanda 595fe2d implements all but the following 3 of your proposals:

* disable gossip No need to run gossip with a single node. Since Cassandra runs in its own deployment, the easiest way to do this is via JMX using the stopGossiping() method on StorageServiceMBean. I need to do some testing with this to see how it works. It looks like the call has to happen after some initialization has completed.

* disable compression  Compression is enabled by default and is configured per table. It reduces disk space but increases CPU utilization. 
ALTER TABLE my_table WITH compression = { ’sstable_compression’: ‘’ };

* Use shared driver instance

Besides, 595fe2d sets `HEAP_NEWSIZE` and `MAX_HEAP_SIZE` env vars so that the embedded cassandra can see them: https://github.com/hawkular/hawkular-commons/pull/56/commits/595fe2dec3daba0f9636148807e434b47051aadd#diff-5c22a8c7340bc3119e2ecaee92e4d375R32

ppalaga:
@jsanda b48778e replaces 595fe2d removing the `HEAP_NEWSIZE/MAX_HEAP_SIZE` stuff. Is it good to merge now?

PR #60, opened 2016-04-21
Improve the swagger documentation on inventory paths.

metlos:
The serialized form of these is always a string with given format
so rather than letting swagger expose the inner structure of these
classes (which doesn't give the end user any meaningful information)
this describes the serialized form of the path in the
swagger-digested annotation.

### DIFF
--- a/hawkular-inventory-paths/pom.xml
+++ b/hawkular-inventory-paths/pom.xml
@@ -28,6 +28,13 @@
 
   <dependencies>
     <dependency>
+      <groupId>io.swagger</groupId>
+      <artifactId>swagger-annotations</artifactId>
+      <version>${version.io.swagger}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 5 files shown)


metlos:
@ppalaga could you please take a look at this?

hawkular-inventory-paths/pom.xml
```diff
@@ -28,6 +28,13 @@
 
   <dependencies>
     <dependency>
+      <groupId>io.swagger</groupId>
+      <artifactId>swagger-annotations</artifactId>
+      <version>${version.io.swagger}</version>
```
ppalaga: What is the relationship of `io.swagger:swagger-annotations` to `com.wordnik:swagger-annotations` that we manage in parent? https://github.com/hawkular/hawkular-parent-pom/blob/41/pom.xml#L322-L323

Are `io.swagger:swagger-annotations` newer and better? Should current occurrences of `com.wordnik:swagger-annotations` be replaced by `io.swagger:swagger-annotations` ?

metlos: @pavolloffay could you please answer that question? I just used the version of `io.swagger` that you introduced in commit https://github.com/hawkular/hawkular-inventory/commit/5667f30a35fcd5fb4b1d6d43bb00cdc91e64a1ee..

pavolloffay:
from [1]: The com.wordnik package is for the older swagger specifications, 1.x. As of swagger specification 2.0, the package is now io.swagger.

[1]: http://stackoverflow.com/a/36359949/4158442

ppalaga:
Replaced by https://github.com/hawkular/hawkular-commons/pull/61

PR #62, opened 2016-05-03
Add Bus specific standalone.xml config, improve nest logging

ppalaga:


ppalaga:
@jpkrohling could you please review and eventually merge? These are things that we had in Accounts and we need to move them to nest now that we are removing the dependencies on Accounts from Hawkular components.

jpkrohling:
Looks good to me!

PR #63, opened 2016-05-06
Use single project name style in all subprojects

ppalaga:


ppalaga:

As for Command Gateway, this is the same code as in https://github.com/hawkular/hawkular-command-gateway/pull/58 

Other commits are just cosmetics: 

* jUnit -> TestNG so that the tests here and in cmdgw use the same framework
* Move Bus integration tests to Nest - to speed up the builds, particularly so that we start the test container only once when building commons
* The rest should be self-explanatory

hawkular-command-gateway/hawkular-command-gateway-api/README.adoc
```diff
@@ -0,0 +1,27 @@
+= Hawkular Feed Communications API
+
+This is used for the server to communication with its feeds over WebSockets. It also is used for UI clients to talk to the server as well (presumably for the UI to manage remote resources that are managed by feeds).
```
jpkrohling: s/to communication/to communicate/

hawkular-command-gateway/hawkular-command-gateway-api/src/main/java/org/hawkular/cmdgw/api/ApiDeserializer.java
```diff
@@ -0,0 +1,156 @@
...
+     * prefixes that JSON with additional information that a Hawkular client will
+     * need to be able to deserialize the JSON.
+     *
+     * This string can be used to deserialize the object via {@link #deserialize(String)}.
+     *
+     * @param msg the message object that will be serialized into JSON
+     * @return a string that includes the JSON that can be used by other Hawkular endpoints to deserialize the message.
+     */
+    public static String toHawkularFormat(BasicMessage msg) {
+        return String.format("%s=%s", msg.getClass().getSimpleName(), msg.toJSON());
```
jpkrohling: Is `msg` guaranteed to not be null?

hawkular-command-gateway/hawkular-command-gateway-api/src/main/java/org/hawkular/cmdgw/api/GenericErrorResponseBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try (PrintWriter pw = new PrintWriter(baos)) {
+            t.printStackTrace(pw);
+        }
+        setStackTrace(baos.toString());
+        return this;
+    }
+
+    private Authentication populateAuthentication() {
+        Authentication auth = this.response.getAuthentication();
```
jpkrohling: Is this still being used? I think there will be no more `authentication` property on the messages, right? Removal of the auth might be in a future PR, so, feel free to ignore this.

hawkular-command-gateway/hawkular-command-gateway-api/src/test/java/org/hawkular/cmdgw/api/ApiDeserializerTest.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    @Test
+    public void testMessageWithLargeExtraData() throws Exception {
+        // serialize a message and some extra data
+        final String testMessage = "this is the message";
+        final String testExtraData;
+
+        // stream [0, 1, 2, ..., 255], repeat that 10 times
+        byte[] bytes = new byte[0xff * 10];
```
jpkrohling: Why not just `new byte[2550]`?

jpkrohling: By the way, isn't there an off-by-one error ? `[0,...,255]` has a length of 256, no?

hawkular-command-gateway/hawkular-command-gateway-api/src/test/resources/log4j.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+    Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
```
jpkrohling: O_o
2014?

hawkular-command-gateway/hawkular-command-gateway-api/src/main/java/org/hawkular/cmdgw/api/GenericErrorResponseBuilder.java
jmazzitelli: we still need SOME kind of auth. We can't allow just anyone to send in commands to stop an app server :) So we need to keep the authentication stuff here - how it actually gets done now that accounts is done will be an exercise for the future.

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/bus/BusEndpointProcessors.java
```diff
@@ -0,0 +1,282 @@
...
+                    // In the future, if we need other queues/topics that need to be listened to, we add them here.
+                    final Endpoint endpoint = Constants.FEED_COMMAND_QUEUE;
+                    BasicMessageListener<BasicMessage> busEndpointListener = new FeedBusEndpointListener(session, key,
+                            endpoint);
+                    return new BusWsSessionListener(Constants.HEADER_FEEDID, key, endpoint, busEndpointListener);
+                }
+            };
+            wsEndpoints.getFeedSessions().addWsSessionListenerProducer(feedSessionListenerProducer);
+
+            uiClientSessionListenerProducer = new BiFunction<String, Session, WsSessionListener>() {
```
jpkrohling: This can be simplified with lambdas...
```
uiClientSessionListenerProducer = (key, session) -> {
  final Endpoint endpoint = Constants.UI_COMMAND_QUEUE;
  BasicMessageListener<BasicMessage> busEndpointListener = new UiClientBusEndpointListener(commandContextFactory, busCommands, endpoint);
  return new BusWsSessionListener(Constants.HEADER_UICLIENTID, key, endpoint, busEndpointListener);
}
```

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/bus/BusEndpointProcessors.java
```diff
@@ -0,0 +1,282 @@
...
+    /**
+     * This creates the bi-function listener-producers that will create listeners which will
+     * create JMS bus listeners for each websocket session that gets created in the future.
+     *
+     * @param ignore unused
+     */
+    public void initialize(@Observes @Initialized(ApplicationScoped.class) Object ignore) {
+        log.debugf("Initializing [%s]", this.getClass().getName());
+        try {
+            feedSessionListenerProducer = new BiFunction<String, Session, WsSessionListener>() {
```
jpkrohling: This can be simplified with lambdas. See the next comment.

hawkular-command-gateway/hawkular-command-gateway-api/src/main/java/org/hawkular/cmdgw/api/GenericErrorResponseBuilder.java
jpkrohling: True, but this auth is not inside the messages. Sockets are upgraded from regular HTTP connections, so, Auth happens already at the HTTP level before the sockets are actually opened.

We had this auth here for two (connected) reasons: sockets can be long lived and OAuth tokens have expiration dates. So, we needed to check on a per-message basis if the token that opened the connection is still valid. We don't need to do that anymore.

What we would need, in the future, is a way to ensure that the user who opened the socket is still valid. I discussed that with Peter earlier this week, and the simplest and most effective way of doing that is to make clients to reconnect after N minutes/hours and to enforce the server to close connections older than N. But this is a problem that can be solved later, as this is to only cover corner cases.

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/ws/ResourcePathDestinationWsCommand.java
```diff
@@ -0,0 +1,76 @@
...
+            ProducerConnectionContext pcc = ccf.createProducerConnectionContext(endpoint);
+
+            MessageId mid = new MessageProcessor().send(pcc, message,
+                    Collections.singletonMap(Constants.HEADER_FEEDID, feedId));
+            log.debugf("Message [%s] forwarded to bus endpoint [%s]", request.getClass().getName(),
+                    endpoint.getName());
+            GenericSuccessResponse response = new GenericSuccessResponse();
+            response.setMessage("The request has been forwarded to feed [" + feedId + "] (MessageId=" + mid + ")");
+            BasicMessageWithExtraData<GenericSuccessResponse> result = new BasicMessageWithExtraData<>(response, null);
+            new WebSocketHelper().sendSync(context.getSession(), result);
```
jpkrohling: Would it make sense to provide a `sendAsync` static method? This seems like a waste of time to create an object without necessity.

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/ws/WsCommand.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * An command to execute on messages coming over a WebSocket.
+ * <p>
+ * Note on thread safety: implementations of this interface must not store any state outside of the
+ * {@link #execute(BasicMessageWithExtraData, WsCommandContext)} method so that the instances can be safely shared
+ * between multiple threads.
+ *
+ * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+ * @param <M> the type of the message to process
```
jpkrohling: Random thought: why do we name type parameters with single letters? We wouldn't accept a code with `void execute(BasicMessageWithExtraData<M> m, WsCommandContext c)`...

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/ws/WsCommands.java
```diff
@@ -0,0 +1,84 @@
...
+    private final EchoCommand echoCommand = new EchoCommand();
+    private final UiSessionDestinationWsCommand uiSessionDestinationWsCommand = new UiSessionDestinationWsCommand();
+    private final EventDestinationWsCommand eventDestinationWsCommand = new EventDestinationWsCommand();
+
+    /**
+     * Returns a collection of {@link WsCommand}s that should handle the given {@code requestClass}.
+     *
+     * @param requestClass the type of a request for which a processing {@link WsCommand} should be found by this method
+     * @return a collection of {@link WsCommand} objects, never {@code null}
+     * @throws NoCommandForMessageException if no {@link WsCommand} was found
```
jpkrohling: No javadoc for the type parameter?

jpkrohling:
Looks good to me. Reviewed all commits, except the 375d5793e29d958ddab104005f2b5c0add441d75 .

ppalaga:
@jpkrohling could you please throw a final look on this? c66aa31 adds some feature-pack related cosmetics and most importantly, it increases the amount of memory for the itest container https://github.com/hawkular/hawkular-commons/pull/63/commits/c66aa31c57532056d354a33ad2ceb2f7aafef4be#diff-5c22a8c7340bc3119e2ecaee92e4d375R197 which probably causes Travis to pass.

jpkrohling:
Github offered me to view only the changes made after my last review. It looks good :)

PR #64, opened 2016-05-13
Put the the Command Gateway tests into a separate module so that they can be re-

ppalaga:


ppalaga:
@jmazzitelli could you please review and eventually merge? This turned out to be necessary for Agent.

PR #65, opened 2016-05-17
Remove credentials from websocket requests as we did before  This is for securit

ppalaga:


ppalaga:

80d715c was an attempt to find the silver bullet that used to be there in Accounts and made the JNDI names to be bound before they were needed by `@Resource` s in `command-gateway.war`. With 80d715c, the issue has not stopped appearing and I turned over to another strategy: the lazy JNDI lookup implemented in 6a35c1a which solved it finally. I admit it is not elegant but it works and I do not have a better idea.

82dc4d6 "Create the itest user during the build with a random password" is there first, to ensure that we do not rely on hard coded defaults too much and second to make it possible to run the itests from cmdgw also in Agent safely. FWIW, the strategy of passing the test user data over the system props has brough some nice improvements in Agent.

4faa335 "Remove credentials from websocket requests as we did before" is not new and you'll surely approve it.

3121e00 is just cosmetics.

hawkular-command-gateway/hawkular-command-gateway-itest/src/test/java/org/hawkular/cmdgw/ws/test/AbstractCommandITest.java
```diff
@@ -39,6 +39,7 @@
         int httpPort = portOffset + 8080;
         baseGwUri = "ws://" + host + ":" + httpPort + "/hawkular/command-gateway";
         authentication = "{\"username\":\"" + testUser + "\",\"password\":\"" + testPasword + "\"}";
+        // System.out.println("testPasword = "+ testPasword);
```
jpkrohling: Did you forget to remove this line, or is it here to make it easier to debug locally?

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/bus/BusEndpointProcessors.java
```diff
@@ -222,8 +221,8 @@ protected void onBasicMessage(BasicMessageWithExtraData<BasicMessage> messageWit
     @Inject
     private BusCommandContextFactory commandContextFactory;
 
-    @Resource(name = Constants.CONNECTION_FACTORY_JNDI)
-    private ConnectionFactory connectionFactory;
+    @Inject
+    private BusConnectionFactoryProvider connectionFactoryProvider;
```
jpkrohling: You could actually inject the ConnectionFactory directly, using a CDI qualifier, like, `@Inject @Ready ConnectionFactory connectionFactory` and have your `BusConnectionFactoryProvider#getConnectionFactory` to be a `@Produces @Ready`. 

But please, use a better name than `@Ready` :)

The qualifier can be something similar to this: https://github.com/hawkular/hawkular-accounts/blob/master/api/src/main/java/org/hawkular/accounts/api/NamedSetting.java

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/bus/BusConnectionFactoryProvider.java
```diff
@@ -0,0 +1,96 @@
...
+import org.hawkular.cmdgw.Constants;
+import org.hawkular.cmdgw.log.GatewayLoggers;
+import org.hawkular.cmdgw.log.MsgLogger;
+
+/**
+ * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+ */
+@ApplicationScoped
+@Default
+public class BusConnectionFactoryProvider {
```
jpkrohling: It's still not clear to me why you need this. Will come back to this later.

hawkular-command-gateway/hawkular-command-gateway-itest/src/test/java/org/hawkular/cmdgw/ws/test/AbstractCommandITest.java
ppalaga: I think I indeed left it there for local debugging.

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/subsystem-templates/hawkular-nest-ejb3.xsl
```diff
@@ -0,0 +1,42 @@
...
+
+  <xsl:output method="xml" version="1.0" encoding="UTF-8" indent="yes" xalan:indent-amount="4" standalone="no" />
+
+  <!-- //*[local-name()='config']/*[local-name()='supplement' and @name='default'] is an xPath's 1.0
+       way of saying of xPath's 2.0 prefix-less selector //*:config/*:supplement[@name='default']  -->
+  <xsl:template match="//*[local-name()='config']/*[local-name()='supplement' and @name='default']">
+    <xsl:copy>
+      <xsl:apply-templates select="node()|comment()|@*" />
+      <!-- Copy the MDB replacement from full to default -->
+      <xsl:copy-of select="//*[local-name()='config']/*[local-name()='supplement' and @name='full']/*[local-name()='replacement' and @placeholder='MDB']"/>
```
jpkrohling: Is this still needed? Do you remember why we had to have this?

hawkular-nest/hawkular-nest-itest/pom.xml
```diff
@@ -290,23 +294,34 @@
       </plugin>
 
       <plugin>
-        <artifactId>maven-resources-plugin</artifactId>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
         <executions>
           <execution>
-            <id>copy-resources</id>
+            <id>create-itest-users</id>
```
jpkrohling: Why do you have a random password for the test user? If you were to use a static password, you wouldn't need to run this, right? You'd just place the appropriate `application-users.properties` and `application-roles.properties` on the WF to be tested.

hawkular-command-gateway/hawkular-command-gateway-itest/src/test/java/org/hawkular/cmdgw/ws/test/AbstractCommandITest.java
jpkrohling: Ok, then just leave a friendly comment above it. Otherwise someone will eventually just remove it :)

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/bus/BusEndpointProcessors.java
ppalaga: > You could actually inject the ConnectionFactory directly

I am akin to think that it would bring the original problem of the ConnectionFactory not being ready under the JNDI name on time. I think so, because `BusEndpointProcessors` is `@ApplicationScoped` and `@Observes @Initialized(ApplicationScoped.class)`. Due to that, the attempt to resolve the JNDI name would happen (1) much earlier than in the present implementation and (2) the `BusConnectionFactoryProvider.getConnectionFactory()` may bolck for some time and I am not sure how much the CDI would accept such a potentially long blocking. WDYT?

jpkrohling: I'm not aware of any limits from CDI. My guess is that it would just wait. I had something similar for the Cassandra `Session` on Accounts, and I never had issues. 

But actually, you might have just found the original issue: the usage of `@Initialized` is new, isn't it? Is it possible that CDI's `@Initialized` event makes this method trigger far before than the previous implementation? How was it previously done? Was it with `@Singleton @Startup` EJB?

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/subsystem-templates/hawkular-nest-ejb3.xsl
ppalaga: You had this in Accounts. That's all I can say. I have no idea why it was there in Accounts.

hawkular-nest/hawkular-nest-itest/pom.xml
ppalaga: > Why do you have a random password for the test user?

As I said, first, to ensure that we do not rely on hard coded defaults too much and second to make it possible to run the itests from cmdgw also in Agent safely (where the passwords may be different).

To put it on other words, with the static `*.properties` files, one stores the same piece of info (username and password) on at least two places: (1) encoded in the `*.properties` file and plain either in pom or in the tests. With the present implementation there is just a single place where the username and password pair is stored: in the `pom.xml` file and it is distributed everywhere where it is needed from there.

And having random passwords makes the test system less vulnerable :)

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/subsystem-templates/hawkular-nest-ejb3.xsl
jpkrohling: So, I tracked this change. Originally, this is what we had: 
https://git.io/vrC6E

And then became this:
https://git.io/vrC61

I don't remember about a change from "full" to "default", though. I also don't think any other component uses MDB, so, I'd just try to remove this. One less thing to maintain.

hawkular-nest/hawkular-nest-itest/pom.xml
jpkrohling: > As I said, first, to ensure that we do not rely on hard coded defaults too much

Ok, now I see what you mean there. 

I'm not sure I see the benefit, to be honest. I usually prefer to have the test to be explicit, ie, setting the username/password itself, rather then getting this value from "somewhere" (the pom or sysprop, in this case).

But I guess this is a matter of taste, so, I'm with this change.

> And having random passwords makes the test system less vulnerable :)

:)

hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/bus/BusEndpointProcessors.java
ppalaga: > How was it previously done? Was it with @Singleton @Startup EJB?

There was no change in this area of Command Gateway since a couple of months. The only difference that occurred recently is that Accounts have gone away. Maybe Accounts just caused some lucky coincidence (e.g. a mere delay by deploying accounts.war before cmdgw.war) that prevented this from happening, because as I said, usually the message in container log reporting that the JMS objects were bound to the respective names appears immediately after the NamingException thrown from the cmdgw deployment.

Indeed, it looks like CDI does not grant anything about the order of an `@ApplicationScoped` bean's initializer and the availability of the `@Resource` s. I have no idea if `@Singleton @Startup` gives such guarantees. 

> I'm not aware of any limits from CDI. My guess is that it would just wait.

I think that what you propose would make the deployment potentially slower because of the waiting during the injection that happens during the deployment. My implementation eliminates the waiting as far as possible through querying JNDI lazily as late as possible - i.e. when the ConnectionFactory is actually used on a client's connect. I admit the present solution is not elegant but it works well enough and I prefer not to spend more time on this.

jpkrohling: Sure, makes sense then. You could also just wrap it on an `Instance`, so that consumers would call `#get()` and CDI resolution would happen then. In any case, I'm OK with this code and this can be refactored later if problems arise. Perhaps then just add a comment on the code pointing to this discussion here?

ppalaga: `Instance` sounds as a good idea, that I'd gladly implement if I only had time for running Agent and Inventory itests enough times to make sure that it works.

hawkular-nest/hawkular-nest-feature-pack/src/main/xsl/subsystem-templates/hawkular-nest-ejb3.xsl
ppalaga: > I don't remember about a change from "full" to "default", though. 

`hawkular-nest-ejb3.xsl` is a copy of `hawkular-accounts-ejb3.xsl` https://github.com/hawkular/hawkular-accounts/blob/master/feature-pack/src/main/xsl/subsystem-templates/hawkular-accounts-ejb3.xsl#L31-L32

`hawkular-accounts-ejb3.xsl` was doing exactly the same thing as https://git.io/vrC6E and https://git.io/vrC61 were doing before we had `hawkular-accounts-ejb3.xsl`.

> I also don't think any other component uses MDB, so, I'd just try to remove this. One less thing to maintain.

No, a quick search for `@MessageDriven` in Hawkular components shows occurences in Alerts, BTM, avail-creator and hawkular-api. I vote for leaving `hawkular-nest-ejb3.xsl` as is.

ppalaga:
@jpkrohling c54bd0d add the comments proposed by you.

ppalaga:
@jpkrohling b910cf2 removes some `println()` s. Good to merge?

jpkrohling:
LGTM

PR #68, opened 2016-05-19
Allow sending an explicit close using TestWebSocketClient

ppalaga:


ppalaga:

jpkrohling:
Is it only a change in the tests? No need to change the actual counterpart for these socket connections?

jpkrohling:
If so, looks good to me.

ppalaga:
> No need to change the actual counterpart for these socket connections?

Let's see what it does like this. We upgraded okhttp, so hopefully setting explicit timeouts there will be enough.

PR #72, opened 2016-06-07
HWKALERTS-148 HWKINVENT-184 - Common Tenant Filter

jpkrohling:
And skip it for the Status filter

### DIFF
--- a/hawkular-rest-status/pom.xml
+++ b/hawkular-rest-status/pom.xml
@@ -47,6 +47,11 @@
   </dependencyManagement>
 
   <dependencies>
+    <dependency>
+      <groupId>org.hawkular.commons</groupId>
+      <artifactId>hawkular-tenant-jaxrs-filter</artifactId>
+      <version>0.7.3.Final-SNAPSHOT</version>
+    </dependency>
 
     <dependency>
       <groupId>javax.enterprise</groupId>
--- a/hawkular-tenant-jaxrs-filter/pom.xml
+++ b/hawkular-tenant-jaxrs-filter/pom.xml
@@ -1,0 +1,146 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
+    and other contributors as indicated by the @author tags.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <artifactId>hawkular-commons-parent</artifactId>
+    <groupId>org.hawkular.commons</groupId>
+    <version>0.7.3.Final-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>hawkular-tenant-jaxrs-filter</artifactId>
+
+  <name>Hawkular REST Filter - Tenant</name>
+  <description>
+    A reusable implementation of a filter that makes Hawkular-Tenant required.
+    Provides also a @OptionalTenant so that endpoints can mark that they don't require a tenant.
+  </description>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.wildfly.bom</groupId>
+        <artifactId>wildfly-javaee7</artifactId>
+        <version>${version.org.wildfly.bom}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+
+      <dependency>
+        <groupId>org.jboss.arquillian</groupId>
+        <artifactId>arquillian-bom</artifactId>
+        <version>${version.org.jboss.arquillian}</version>
+        <scope>import</scope>
+        <type>pom</type>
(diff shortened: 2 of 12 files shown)


jpkrohling:
@theute , @jshaughn , could you please review? I tested this with my local changes to Alerts, which will follow next once this is merged.

jshaughn:
LGTM, @theute can merge unless he sees something...

jpkrohling:
As @theute is on PTO, I'm merging this. If there are any issues, we can either revert or fix it.

hawkular-tenant-jaxrs-filter/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+    <groupId>org.hawkular.commons</groupId>
+    <version>0.7.3.Final-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>hawkular-tenant-jaxrs-filter</artifactId>
+
+  <name>Hawkular REST Filter - Tenant</name>
+  <description>
+    A reusable implementation of a filter that makes Hawkular-Tenant required.
+    Provides also a @OptionalTenant so that endpoints can mark that they don't require a tenant.
```
abonas: I didn't find this part as part of this patch

jpkrohling: Nice catch! I renamed it to `@TenantRequired(boolean)` , which sounded better. I'll send a PR with the fix.

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantFeature.java
```diff
@@ -0,0 +1,49 @@
...
+    public void configure(ResourceInfo resourceInfo, FeatureContext context) {
+        Class<?> resourceClass = resourceInfo.getResourceClass();
+        Method method = resourceInfo.getResourceMethod();
+
+        boolean required = true;
+        if (resourceClass.isAnnotationPresent(TenantRequired.class)) {
+            required = resourceClass.getAnnotation(TenantRequired.class).value();
+        }
+
+        if (method.isAnnotationPresent(TenantRequired.class)) {
```
abonas: from how I understand the code, the class could have an annotation with value "true", but a method in the same class could have the annotation with the value "false" that will reset required to "false" - is that how it's supposed to work? (just making sure)

jpkrohling: Yes. That's for the case where all methods on a class don't require the Hawkular-Tenant header, except one: instead of marking all methods with "false", the class is marked with "false" and a single method is marked with "true". This behavior is in line with similar Java EE annotations: the most specific setting is always the one which is effective.

hawkular-rest-status/pom.xml
```diff
@@ -47,6 +47,11 @@
   </dependencyManagement>
 
   <dependencies>
+    <dependency>
+      <groupId>org.hawkular.commons</groupId>
+      <artifactId>hawkular-tenant-jaxrs-filter</artifactId>
+      <version>0.7.3.Final-SNAPSHOT</version>
```
metlos: We shouldn't depend on snapshots.

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantFeature.java
```diff
@@ -0,0 +1,49 @@
...
+            required = method.getAnnotation(TenantRequired.class).value();
+        }
+
+        if (required) {
+            context.register(new TenantFilter());
```
metlos: The tenant filter is stateless.. Wouldn't it be better to not create a new instance of it on every (tenant-requiring) request?

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantRequired.java
```diff
@@ -0,0 +1,38 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import javax.enterprise.util.Nonbinding;
+import javax.ws.rs.NameBinding;
+
+/**
+ * @author Juraci Paixão Kröhling
+ */
+@NameBinding
```
metlos: Is `@NameBinding` required if the need for the filter is determined by the dynamic feature?

jpkrohling: +1 , on a previous iteration, the filter was a JAX-RS filter annotated with `@Provider`. As that's not the case anymore, I'll remote the `@NameBinding`

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantFeature.java
jpkrohling: +1

hawkular-rest-status/pom.xml
jpkrohling: Maven manages this just fine and will update it on releases. Using `${project.version}` used to cause problems when the build is "embedded" into another build (release builds). It has been fixed for some time, but I believe the best practice is still not to use `${project.version}`.

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantFeature.java
metlos: While it might not be the best Maven practice (I honestly don't know and believe your expertise here), it's been used in Hawkular all over the place:
`find -name 'pom.xml' | xargs grep '${project.version}' | wc -l` in various components:

metrics: 42
inventory: 60
alerts: 33

hawkular-rest-status/pom.xml
metlos: While it might not be the best Maven practice (I honestly don't know and believe your expertise here), it's been used in Hawkular all over the place:
`find -name 'pom.xml' | xargs grep '${project.version}' | wc -l` in various components:

metrics: 42
inventory: 60
alerts: 33

jpkrohling: In the name of consistency, I changed it :)

PR #73, opened 2016-06-08
Fixed annotation name on pom and added some javadoc

jpkrohling:


jpkrohling:
Follow up on #72

jpkrohling:
@abonas , this should fix the comments from #72 . Would you please review/merge, once Travis approves it?

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantFeature.java
```diff
@@ -24,6 +24,9 @@
 import javax.ws.rs.ext.Provider;
 
 /**
+ * A JAX-RS dynamic feature that adds the {@link TenantFilter} to all methods which have {@link TenantRequired}
+ * effectively as "true".
```
abonas: when TenantRequired is not specified at all, from https://github.com/hawkular/hawkular-commons/pull/72/files#diff-112e3d7f5467ae0e9ba60c0a73f8f51eR36 I understand it is still considered as required. If that's the case, the above explanation can be misleading to think that TenantRequired must be declared.

jpkrohling:
@metlos , @abonas , I applied the changes from the review of this PR and #72 . Would you please review? The only pending change is on `-SNAPSHOT` that @metlos mentioned on #72 , and which I'm open to change if @metlos requires it :)

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantFeature.java
```diff
@@ -24,10 +24,16 @@
...
 /**
+ * A JAX-RS dynamic feature that adds the {@link TenantFilter} to all methods which have {@link TenantRequired}
+ * effectively as "true", which might be the default value for {@link TenantRequired}. In the usual case, this feature
+ * is added to all classes and methods, except when classes/methods are annotated with "false".
+ *
  * @author Juraci Paixão Kröhling
  */
 @Provider
 public class TenantFeature implements DynamicFeature {
+    private static final TenantFilter TENANT_FILTER = new TenantFilter();
```
abonas: is this thread safe?

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantRequired.java
```diff
@@ -24,12 +24,14 @@
...
-import javax.ws.rs.NameBinding;
 
 /**
+ * Explicitly marks a class or method on the requirement for the Hawkular-Tenant HTTP header. Method annotations have
+ * priority over class annotations, so, if a method sets this as "true" and the class sets as "false", a Hawkular-Tenant
+ * will be required for calls that end up on that method.
+ *
  * @author Juraci Paixão Kröhling
  */
-@NameBinding
```
abonas: why is this removed?

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantFeature.java
metlos: Yes. Constants are resolved at classloading time (basically in a synthetic static initializer), which is guaranteed to be called exactly once by the JVM.

hawkular-tenant-jaxrs-filter/src/main/java/org/hawkular/jaxrs/filter/tenant/TenantRequired.java
metlos: See the discussion: https://github.com/hawkular/hawkular-commons/pull/72#discussion_r66230053

jpkrohling:
@metlos , I just updated this PR to use ${project.version} instead of the actual version.

metlos:
LGTM.

PR #75, opened 2016-06-09
Fixed encoding on deserializer test.

jpkrohling:


jpkrohling:
On systems without UTF-8 as default, this test fails, as the string is not created with the platform's encoding, but it's read with the platform's encoding. They should all be the same: either platform's, or explicit.

PR #76, opened 2016-06-14
HAWKULAR-1089 - Removed embedded Cassandra

jpkrohling:


jpkrohling:
@pilhuhn , could you please review/merge? Once it's merged, I'll release commons so that it can be incorporated early by other components.

pilhuhn:
Completely removing it from hawkular-commons seems to be wrong.
We must not include it in hawkular-services, but may still want to include it in a hawkular-full distribution

jpkrohling:
Then hawkular-full would have to re-add this then, IMO. The nest is the base for our distribution, so, if we don't want Cassandra at the intermediate distribution, we shouldn't have it at the basic building block.

jpkrohling:
Or did you mean that we should keep the maven module `embedded-cassandra` ? If so, then I agree that we could keep it.

pilhuhn:
I think I meant the maven module :)

jpkrohling:
Ok, we could keep it here, or we could maintain it at the hawkular-all level. I'm not sure I like this module here in Commons, as it implies that other components are free to use it, like any other commons module.

We could decide on that later, though.

jpkrohling:
@pilhuhn : ready for re-review

jpkrohling:
I'll change this PR to have the integration tests to use embedded Cassandra.

jpkrohling:
Updated the commit to use embedded Cassandra on itests.

jpkrohling:
Reviewed via IRC.

PR #77, opened 2016-06-19
@NonBinding is not needed

metlos:
because the `@TenantRequired` is not used for injection or container-managed interception.

Removing it enables us to minimize the dependencies of this module.

### DIFF
--- a/hawkular-tenant-jaxrs-filter/pom.xml
+++ b/hawkular-tenant-jaxrs-filter/pom.xml
@@ -64,8 +64,8 @@
 
   <dependencies>
     <dependency>
-      <groupId>javax</groupId>
-      <artifactId>javaee-api</artifactId>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jaxrs</artifactId>
       <scope>provided</scope>
     </dependency>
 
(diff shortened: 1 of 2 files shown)


jpkrohling:
LGTM

hawkular-tenant-jaxrs-filter/pom.xml
```diff
@@ -64,8 +64,8 @@
 
   <dependencies>
     <dependency>
-      <groupId>javax</groupId>
-      <artifactId>javaee-api</artifactId>
+      <groupId>org.jboss.resteasy</groupId>
```
jpkrohling: Is there a reason to change this dependency? The `javax.javaee-api` should not bring any "extra" dependency that we don't already have provided by the AS.

metlos:
Metrics don't depend on it and I guess it is the reason they freaked out when asked to use this instead of their own tenant handling.

hawkular-tenant-jaxrs-filter/pom.xml
metlos: Metrics don't depend on it and I guess it is the reason they freaked out when asked to use this instead of their own tenant handling.

metlos: They already depend on resteasy so I really do hope we can all standardize on using tenant filter.

jpkrohling: They freaked out for no reason, then. If that helps to standardize on this filter, then so be it. But there's no difference between those two dependencies, technically speaking...

metlos: I agree :wink:

pilhuhn:
Is this good to go?

jpkrohling:
I think so. I was just curious about the choice of the dependencies, which brings no real change.

PR #78, opened 2016-06-27
Hawkular-inventory-paths is provided by the nest.

jshaughn:
The cmd gateway war should not package it, and it should
declare a dependency in jboss-deployment-structure.xml


stefannegrea:
Thanks @jshaughn!

PR #79, opened 2016-07-28
Related to Hawkular-1101, Backfiller does not honor DOWN availability

jshaughn:
For feeds establishing a websocket connection to the server we
want to detect when the WS is closed, and use that information
to backfill avails for the resources managed by that feed. So,
put a message on the bus that can be picked up by a hawkular-level
(aka glue) listener and actioned on.

@jmazzitelli please review


jmazzitelli:
LGTM

PR #81, opened 2016-08-08
HWKAGENT-123 add force deploy and remove content options

jmazzitelli:


jshaughn:
forceDeploy seems OK.  I guess this becomes content replacement.  Could the description indicate what happens if it exists and is currently enabled?  I would guess it does an undeploy, then replaces the content, then applies the enabled flag?  Also, what is the default?

removeContent is sort of OK.  I think this conflicts with the fact that we currently offer both Stop and Undeploy on a Deployment in MIQ.   Stop is currently a passthru to the Deployment's Undeploy operation whereas Undeploy now maps to our UndeployApplicationRequest.   In short, if our Undeploy option did not perform as removeContent=true it would effectively be a Stop.  So that's sort of a no-op.  I assume the default here would be true but it's not in the description.

I think what we need to sort out is what should be offered at the Deployment level and what at the Server level.

jmazzitelli:
> Stop is currently a passthru to the Deployment's Undeploy operation whereas Undeploy now maps > our UndeployApplicationRequest

And this is the problem I think. I don't think we should be passing through because its a problem in domain mode (we need to send this directly to the host controller)

> I think what we need to sort out is what should be offered at the Deployment level
> and what at the Server level.

As per the call we had today, there is nothing at the Deployment level. It is either on the WildFly Server level (standalone) or server group level (domain mode). This is from the viewpoint of the UI.

jshaughn:
> As per the call we had today, there is nothing at the Deployment level. It is either on the WildFly Server level (standalone) or server group level (domain mode). This is from the viewpoint of the UI.

I'm not really seeing anything in the PRD that says this, maybe it needs to be made more clear?  From a UI perspective it probably flows more naturally the way it is today, navigate to a deployment, under the standalone server or perhaps server group or domain, and execute Enable/Disable/Restart/Undeploy.  As long as the operations do the right thing.  To do this directly from the server level would require a deployment list anyway.  For our impl the Undeploy should always be performed using ApplicationUndeployCommand.  This allows the agent to also interact with inventory, and potentially do conditional things depending on the server-level (standalone, domain, etc).  The question for me is whether Enable/Disable/Restart native deployment operations "do the right thing" depending on whether that deployment's parent's server-level.  If not then we'll likely need dedicated command requests for those as well.

PR #83, opened 2016-09-01
Hawkular 1096 ansible commands (do not merge)

rubenvp8510:
This PR is not intended for merge. I want to show a problem that I have trying to get information from the inventory.

One thing I trying to do is: given a resource canonical path through an AnsibleRequest, get the host, bind IP, and WildFly home path and pass it to ansible via extraVars so ansible playbook can access to that information. I'm doing all of this inside the AnsibleCommand.

I tried to use the inventory API on the AnsibleCommand to get Wildfly configurations, but this injects a dependency here and this is bad because this is the common project (all projects including inventory depends on it)

In this line I fetch the information:
https://github.com/rubenvp8510/hawkular-commons/blob/94ec515c22b5dc942e7bdccb3b8cc1904ee07a3d/hawkular-command-gateway/hawkular-command-gateway-war/src/main/java/org/hawkular/cmdgw/command/ws/AnsibleCommand.java#L63

One solution could be doing a separate module for processing the ansible request, and in the command put the request in a queue. that could work?

### DIFF
--- a/hawkular-command-gateway/hawkular-command-gateway-war/pom.xml
+++ b/hawkular-command-gateway/hawkular-command-gateway-war/pom.xml
@@ -89,6 +89,11 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-api</artifactId>
+      <version>0.17.1.Final</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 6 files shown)


hawkular-command-gateway/hawkular-command-gateway-war/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -24,6 +24,7 @@
       <module name="org.hawkular.bus" />
       <module name="org.hawkular.commons.cassandra-driver" />
       <module name="org.hawkular.commons.hawkular-inventory-paths"/>
+      <module name="deployment.hawkular-inventory-dist.war" services="import"/>
```
jmazzitelli: I definitely think we do not want to pull this in. This is the full inventory war I think. Not really what we want.

hawkular-command-gateway/hawkular-command-gateway-war/pom.xml
```diff
@@ -89,6 +89,11 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-api</artifactId>
```
jmazzitelli: is the other inventory dep not sufficient? (inventory-paths). I am guessing it is not. We should ask Lukas about splitting out another inventory library if we need it.

PR #84, opened 2016-09-15
Moved hawkular-inventory-api from inventory to commons.

rubenvp8510:


pom.xml
```diff
@@ -100,13 +101,21 @@
...
     <version.org.wildfly.arquillian>2.0.0.Alpha1</version.org.wildfly.arquillian>
+    <version.io.reactivex>1.0.8</version.io.reactivex>
 
     <plugin.version.org.jsonschema2pojo>0.4.14</plugin.version.org.jsonschema2pojo>
   </properties>
 
   <!-- We manage the artifacts from Commons here so that hawkular-commons-parent can be imported as a BoM by downstream Hawkular components -->
   <dependencyManagement>
     <dependencies>
+      <dependency>
```
metlos: This dependency doesn't seem needed for the inventory API itself. It might be for the CDI that I believe you're planning on moving over, too, though?

rubenvp8510: I put this here because  I saw that inventory API uses  jboss-logging and I imported this pom because it has the artifact version.

rubenvp8510:
I'm having compilation problems on Travis, but not locally, It fails to compile the inventory-API, I clean my repositories and tried again and still works locally but not on Travis.

It seems like it is Travis JDK version but I'm not totally sure ( hawkular-inventory Travis has a different version)

WDYT @metlos  @jmazzitelli  Have you ever seen this issue before?

metlos:
@rubenvp8510, yes you need a custom JDK version - a newer one than the default one on Travis. Look at `.travis.yml` of the inventory repo on how it's done.

jmazzitelli:
why do you keep closing and reopening  PRs?

rubenvp8510:
Ups, sorry, it was because I had problems with Travis and I need to re-run the tests again.

rubenvp8510:
btw, I already solve the issues and this PR is ready for another review :)

metlos:
How much of a blocker is this to you? I am now in the process of updating the API with the support for versioning, so this would make my life quite a lot more difficult.

rubenvp8510:
@metlos not too much, this is only for break a compilation dependency between commons and inventory when you need to access to inventory API inside a command, I can work assuming that this movement is gonna happen sooner or later.

stefannegrea:
Why is commons depending or needs to be aware of a consuming service?

rubenvp8510:
@stefannegrea  I need this because in one command on command-gateway I need to fetch some information from inventory. (See this PR #83 )

stefannegrea:
I see Travis infrastructure updates in the same PR. Could you please open another PR with just those changes?

PR #87, opened 2016-10-17
Create slim profile for OpenShift integration.

stefannegrea:


stefannegrea:
All self dependencies are removed since that is not the method used for productization; all those declarations were superfluous.

PR #88, opened 2016-10-18
Fix license header years because subprojects have different inception…

stefannegrea:
… year settings that get ignored when submodules are not part of the reactor. The inception year should only be set at the parent pom of respository to allow selective maven reactor.


jshaughn:
I'l take your word for it :)   It's green so I reckon it's ok. merging...

PR #90, opened 2016-11-09
Implements EventDestination on Add/Remove Datasource  Command

rubenvp8510:
Implements EventDestination interface so it triggers an event creation when someone requests an operation for add or delete Datasources.

@jshaughn I think this was the problem of the BZ 1390756 "No MW Provider Event Shown After Deleting Datasource", Could you review this, please? Thanks :) 

Also, I'm not sure if we should implement this for other commands.


jshaughn:
@rubenvp8510 , excellent catch.  I had forgotten that for the Response to be posted to the "HawkularCommandEvent" queue that the generated class needed to implement that interface.  The interface only causes the Response to be posted to the queue, it does not guarantee an event is generated, we need to have additional support in place for that to happen (in the  listener code in Hservices and in the event catcher code in MIQ).  As it stands now these are the only two that we need to update.  We don't have the additional support in place for any others.

I will merge now.

PR #91, opened 2016-11-11
CORS filter allow empty allowed origins

pavolloffay:
@jpkrohling could you please review?


pavolloffay:
@jshaughn @stefannegrea Could you please review too?

jshaughn:
It's not totally clear to me what value we get from a null/empty CORS filter, but this seems fine for handling the situation.  What is the use case?  If it is to protect against a bad setting perhaps we  should log a warning.

pavolloffay:
Basically, it is to protect against a bad setting. When null was passed it threw NPE, when `""` was passed it threw NPE later in a path evaluation, so this PR should protect against these two errors.

jpkrohling:
Looks good to me.

PR #92, opened 2016-11-15
Introduce mvnw - a way to ensure that devs and CI use the same Maven version  A 

ppalaga:


ppalaga:
This depends on https://github.com/hawkular/hawkular-parent-pom/pull/93

.mvn/srcdeps.yaml
```diff
@@ -0,0 +1,56 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# `.mvn/srcdeps.yaml` - the srcdeps configuration file
+
+# `configModelVersion` exists just in case we'll have to change the schema of srcdeps.yaml in the future.
+#
+# Required, no default, only 1.0 supported at the moment
```
jpkrohling: If only 1.0 is supported, why is the next line 1.1? :)

pom.xml
```diff
@@ -23,7 +23,7 @@
   <parent>
     <groupId>org.hawkular</groupId>
     <artifactId>hawkular-parent</artifactId>
-    <version>49</version>
+    <version>50-SRC-revision-0422f95715bdbc9f91fa343510e032c3b466c9dc</version>
```
jpkrohling: This needs to be updated.

.mvn/srcdeps.yaml
ppalaga: Yes, the comment is out of date. The proper solution would be to put a link to configModelVersion : srcdeps-maven compatibility matrix that I should add to srcdeps-maven docs somewhere. Let me do it before I forget. Perhaps a link to a general srcdeps.yaml doc would be a better choice here. Such one is less probable to get out of date.

jpkrohling: Or just remove it... We need this kind of doc only on the example we use for documentation purposes. Here, it's just a consumer, so, those comments are not needed.

ppalaga: I do think that it will be very handy for the maintainers to have a link to the srcdeps.yaml reference at hand here.

ppalaga: BTW, the CI is passing here now is a proof that the -SRC- version in parent works. Although it is great that it works, I think it will be more practical to release `hawkular-parent` 50 now and upgrade it here straight away in this PR. WDYT? Could you please eventually release hawkular-parent now?

ppalaga:
6c35a50 replaces 4ee0a12 with:

* Upgrade to hawkular-parent 50
* Removed comments from srcdeps.yaml and replaced them with links to the srcdeps documentation: https://github.com/hawkular/hawkular-commons/pull/92/commits/6c35a50452fb043b8e8e8779782dca954e49189d#diff-34efcdaa51afff46bba8cd9c1387e6f5R20

ppalaga:
This is ready to be merged from my PoV

PR #93, opened 2016-12-08
[HWKMETRICS-554] Fix several issues with CORS

jotak:
- payload was sent even on error 400 on invalid origin. This was because CORS validation occured at response filtering => move validation to request filter

- CORS related headers not sent for some endpoints. Related to how NamedDataPointObserver was handling the async response, i think the output stream was closed and subsequent response filter adding headers were ineffective; now use AsyncResponse::resume to complete the response.

Also replace abstract classes with utility methods so that subclasses can have better control on filters (and decrease coupling).


jotak:
For information, I've made changes locally on inventory, alerts and apm to reflect the API change... though I'm not sure if they're all ready to use an upgraded hawkular-commons

jshaughn:
@jotak Now that Alerting is packaged with Metrics, we try to stay on the same Commons version.  So wherever we update Metrics we should do the same for Alerting, whether it be masters or release branches.

jshaughn:
LGTM. Thanks, Joel!

PR #98, opened 2017-04-14
Move inventory data model & mixins to commons

jotak:
It has been cleaned from all unnecessary code.
Both hawkular-agent and hawkular-services will rely on this


jotak:
I think we can put it in commons, since it's used both by the agent and services.

metlos:
Closing this in favor of #99.

PR #99, opened 2017-04-18
Move inventory data model & mixins to commons

metlos:
This adds just one commit to #98 that fixes the test failure introduced by it.


metlos:
Squashed the commits into one as per request.

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/ExtendedInventoryStructure.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Collection;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * @author Joel Takvorian
+ */
+public class ExtendedInventoryStructure {
```
jotak: I was not inspired at all when I found this name. Renaming like "IndexedInventoryStructure" would be more appropriate. But it would need to be propagated to h* agent & services

jotak: [of course I can do it later, out of this PR ;) ]

PR #100, opened 2017-04-24
Move license checker exclusions to apply globally

jshaughn:
and exclusion for .log files because release builds started failing due to files like hs_err_pid22992.log, generated during a test (probably due to openjdk).


jotak:
LGTM

PR #103, opened 2017-05-08
HWKAGENT-177 xa datasource class is obsolete, remove it

jmazzitelli:


abonas:
I'm confused. wasn't this added in https://github.com/hawkular/hawkular-agent/pull/338 ?
why is this obsolete?

jshaughn:
LGTM

jmazzitelli:
@abonas I quickly added both to get @mtho11 going - I wasn't sure which of the two were used.

I found out (see the JIRAs and wildfly-dev thread on this) that Wildfly never uses and should have removed xa-datasource-class. We should not be using it. Thus this removes it.

@mtho11  knows about this - we talked about it last week - and he is not using this prop in the MiQ code.

jmazzitelli:
if you want to read all about this - read https://issues.jboss.org/browse/HWKAGENT-177 and the links within that JIRA. This wildfly-dev post also discusses this: http://lists.jboss.org/pipermail/wildfly-dev/2017-May/005913.html

PR #104, opened 2017-05-30
Extract shared artifacts from Alerting

lucasponce:
Common logging & properties utils
Rest Api for Vert.X/Netty handlers
Common Vert.X/Netty server loader

As commented @stefannegrea. 
Please could you review and merge ?
I have a local alerting branch working with them, I would need a release in orther to link the dependencies (or just a merge with temporal srcdeps).

/cc @jshaughn

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@
   </ciManagement>
 
   <properties>
+    <version.io.vertx>3.4.1</version.io.vertx>
     <version.javax.javaee>7.0</version.javax.javaee>
     <version.com.google.code.findbugs>1.3.9</version.com.google.code.findbugs>
     <version.commons-logging.commons-logging>1.1.1</version.commons-logging.commons-logging>
@@ -90,6 +91,10 @@
     <version.org.wildfly.arquillian>2.0.0.Alpha1</version.org.wildfly.arquillian>
 
     <plugin.version.org.jsonschema2pojo>0.4.14</plugin.version.org.jsonschema2pojo>
+
+    <!-- Used for utils. Utils is used without Widlfly depenencies -->
+    <version.org.jboss.logging>3.3.1.Final</version.org.jboss.logging>
+    <version.org.jboss.logging.jboss-logging-tools>2.0.2.Final</version.org.jboss.logging.jboss-logging-tools>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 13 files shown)


lucasponce:
In this local WIP branch
https://github.com/lucasponce/hawkular-alerts/tree/next
you can see how itests are working using the shared artifacts.

jshaughn:
@lucasponce This all looks good to me, I think.  The short module naming is different than what is there already but I'm not opposed to it.  I like moving towards a consolidated properties file.  The term HawkularServer could maybe get confusing with Hawkular Services but it makes sense and I don't have a better name.  I think we should maybe hold on a release but I'm against a merge and the use of a src dep.  The other alternative would be be to create a "next" branch for commons as well, but that may be unnecessary.  I guess the main question is whether there is buy-in for vert.x stuff in commons.

lucasponce:
I am afraid that for travis uses and several repos we need to rely on srcdeps or releases.
@stefannegrea commented that it would be good to use commons for shared artifacts, I think we can move further and refine in iterative steps.

jshaughn:
LGTM, @stefannegrea, any objections to merge?

stefannegrea:
@lucasponce, all these new modules should be placed under a folder since they are specific to a new rest implementation.

server/src/main/java/org/hawkular/HandlersManager.java
```diff
@@ -0,0 +1,136 @@
...
+                                RestEndpoint endpoint = (RestEndpoint)clazz.getAnnotation(RestEndpoint.class);
+                                for (int j=0; j<interfaces.length; j++) {
+                                    if (interfaces[j].equals(RestHandler.class)) {
+                                        endpointsClasses.put(endpoint.path(), clazz);
+                                    }
+                                }
+                            }
+                        } catch (Exception e) {
+                            log.errorf(e, "Error loading Handler [%s].", className);
+                            System.exit(1);
```
lucasponce: At this point yes. As it sounds a fatal error (i.e. corrupted file or similar).
Another thing is if a handler has some runtime error which is handled in the start() and error in logged but no hard stop.
(Good thing of this approach is this logic can be revisit if we thing it should be different).

lucasponce:
@stefannegrea comments addressed, thanks for the review.

lucasponce:
pom.xml fixed

jshaughn:
I think the comments are handled, I'll merge so that @lucasponce can set a srcdep in his pending PR for alerts.  If necessary, further changes can be handled as add-ons.

PR #105, opened 2017-06-16
Update the handler manager to allow handlers in subpackages and also …

stefannegrea:
…parse classloader resources for handlers.


lucasponce:
LGTM
Thanks @stefannegrea

PR #106, opened 2017-06-19
Filter correctly handlers on *hawkular*.jar files

lucasponce:
Fix a bug in the scanner of handlers as if hawkular is part of a directory a third party jar passed the filter and it is processed for scanning (which is something we don't want).
So, filter is limited to the jar name, not the full path.


jshaughn:
LGTM

PR #108, opened 2017-06-28
HAWKULAR-1233 Create AvailChange notification

josejulio:


jshaughn:
@josejulio This looks good.  I can merge it if it helps you but seeing that it is just one part of a potential solution we may want to wait until you see whether the entire approach works as we hope.  Let me know how you want to handle it.

josejulio:
I prefer to wait until we have everything working together

josejulio:
After checking the agent code, not sure if we would like to differentiate the notifications that apply to the metric-set-dmr [1] and those who apply to the avail-set-dmr [2]. So that you aren't allowed to use a server-added on a avail and avail-change on a regular metric.

This would require e.g. making the Notification class use a generic. So we could plug a NotificationTypeMetric or NotificationTypeAvail.

WDYT?

[1] https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml#L42
[2] https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml#L540

jshaughn:
@josejulio I see your point but I wouldn't worry about it too much, I'd just let the agent config editor "do the right thing".  I think I'd just put all of the notifications at the type-level and let them apply where it makes sense.  We should get an opinion from @jmazzitelli.

josejulio:
@jshaughn I added an availType property to the `AVAIL_CHANGE`.
On the current configuration we only have at most one avail per resource, but it can have more. Having the name will prevent shadowing on different avails of the same resource.

jshaughn:
@josejulio, If I understand you I think you are saying that the AVAIL_CHANGE should report the previous and new avail.  That makes sense.  Although minor, I'd like to see  "availType" and "newAvail" use the same naming, like "Avail" and "newAvail" or "availType" and "newAvailType".

josejulio:
Sorry, I mean that each resource can have more than one avail.
So I need to be able to tell which Avail is the one that changed. That's the availType for.
e.g.
availType = 'Server Availability' [1]

If in the future we add other availability to the set, it would be good to not mix them.

[1] https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml#L543

jshaughn:
Why would any one resource have multiple avail metrics?  It can only have one avail value at any one time, no?

jmazzitelli:
It is technically possible to have multiple avail metrics - nothing in the code would prevent this. You can have multiple avails just like you can have multiple metrics. Under what scenarios would that make sense, I dunno, but it is at least possible if you want to do it.

jshaughn:
I agree, but it seems like technical excess.  Not a big deal, we can leave this as is.

jshaughn:
LGTM, Merging this. We can generate a release or you can update the dependent PRs with a src-dep.

PR #109, opened 2017-06-28
Restrict resource loading to only Hawkular related classes from jars. …

stefannegrea:
…Also capture errors when trying to load a class.


lucasponce:
LGTM

PR #110, opened 2017-06-29
Skeleton for remote HawkularServer MBean invocation

lucasponce:
Introduced a HawkularManager utility class for command line invocation of HawkularServer MBean.
It is just a skeleton for status/stop operations but needed to stop properly a server, invoking the chained .stop() calls.
No security implemented yet, it is remote instead of using attach VM utilities to simplify as a) to allow remote operations b) to not deal with the specific of tools.jar per platform.


jshaughn:
LGTM

PR #111, opened 2017-06-30
Support relative endpoints on multiple applications

lucasponce:
This PR fixes how endpoints are internally stored, adding the class into the relative endpoint.
So, multiple relative enpoints on different applications are now supported.

### DIFF
--- a/vertx/server/src/main/java/org/hawkular/HandlersManager.java
+++ b/vertx/server/src/main/java/org/hawkular/HandlersManager.java
@@ -49,8 +49,8 @@ public class HandlersManager {
     private Router router;
     private Map<String, BaseApplication> applications = new HashMap<>();
     private Map<String, Class<BaseApplication>> applicationsClasses = new HashMap<>();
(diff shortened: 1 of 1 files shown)


jshaughn:
LGTM, Just For clarity...
(11:18:15 AM) jshaughn: so what did this mean, it allows the same endpoint from different classes?  
(11:19:27 AM) jshaughn: but when making a REST request how do you choose an endpoint when the path is duplicated?
(11:19:39 AM) jshaughn: or is it the baseUrl will be different?
(11:20:04 AM) jshaughn: like ../metrics/foo and ../alerts/foo
(11:20:12 AM) jshaughn: where /foo is the shared path
(11:22:16 AM) lponce: yes
(11:22:36 AM) lponce: yes ../metrics/foo and ../alerts/foo now are fixed
(11:22:53 AM) lponce: there was a bug in the HandlersManager

jshaughn:
Build passed, travis hung. Merging...

PR #112, opened 2017-07-05
Optional CorsHandler for UI development

lucasponce:
This PR adds a basic CorsHandler into the baseUrl.
During UI work, a local dev uses a different domain to access into the server and Cors rules are necessary.
Perhaps it can also get necessary in other scenarios and we might want to add more options (restrict the headers, methods, etc) but for now I think is good.


jshaughn:
Can we leverage any of what we did with the jaxrs cors filter stuff, also in commons?  Anyway, this seems fine to start.

lucasponce:
Not sure. We don't use jaxrs or we have plans to reuse any *Servlet related technology.

PR #113, opened 2017-07-07
CORS Support

jshaughn:
- Add allowed methods
- change property name from hawkular.cors-url to hawkular.cors-origin-pattern
  - this better reflects what the value actually is. It is a pattern and supports
    this like, .*localhost:.*

I carried over the defaults we used in the jaxrs cors support module
in commons.


jshaughn:
@lucasponce or @jmazzitelli please review.

jmazzitelli:
much nicer. obviously, we need to fix hawkular.properties and perhaps other places in hawkular-alerting due to this change here.

jshaughn:
Yes, assuming this goes in we can update the properties file and the commons src-dep.

lucasponce:
LGTM

PR #114, opened 2017-07-19
Adding property to delete JDR report file after export operation ends

israel-hdez:
Export JDR operation is streaming to the client. It may not make sense
to keep a copy of it in the server node. Adding a property to allow the
client to choose if the report should be kept. Being a boolean variable,
default is to keep a copy.


jshaughn:
LGTM

PR #115, opened 2017-09-12
WIP New inventory model first draft

jotak:
(not definitive)

Move oldmodel to another package, keep it for the time being but it should be totally removed later.
Keep only the necessary parts, formalize more explicitly model

Getters/Setters/Constructors not generated yet.


hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
```diff
@@ -18,19 +18,19 @@
...
-import org.hawkular.inventory.api.model.Feed;
-import org.hawkular.inventory.api.model.MetadataPack;
-import org.hawkular.inventory.api.model.Metric;
-import org.hawkular.inventory.api.model.MetricType;
-import org.hawkular.inventory.api.model.OperationType;
-import org.hawkular.inventory.api.model.Relationship;
-import org.hawkular.inventory.api.model.Resource;
-import org.hawkular.inventory.api.model.ResourceType;
-import org.hawkular.inventory.api.model.Tenant;
+import org.hawkular.inventory.api.oldmodel.AbstractElement;
```
lucasponce: (Forgive me if I make some basic questions, most of them are related I'm not familiar with the details of the inventory project)
Question:
Do we need to main the oldmodel ?

jotak: I can remove it, it's too distracting in code review...

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Metric.java
```diff
@@ -16,175 +16,13 @@
...
-                return new Blueprint(metricTypeId, id, name, collectionInterval, properties, outgoing, incoming);
-            }
-        }
-    }
+public class Metric {
+    private String id;
+    private String name;
+    private String feed;    // Index; But not sure if feeds are still in play if the inventory is built from directly prometheus scans
+    private MetricType type;  // Index [Search all metrics of type xx]
+    private String root; // needed? Check if query exists "select all metrics for root"
```
lucasponce: "root" is having a role of resource parent ? (i.e. this metrics belongs to XYZ ?)

(Again, this is for me, I can merge and ask later, whatever is more comfortable)

jotak: Yes, actually the only role of "root" is to act as an index to search all resources of a given root. Beside that, it's useless.

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Resource.java
```diff
@@ -16,140 +16,21 @@
...
+public class Resource {
+    private String id;  // Unique index [Search resource by id]
+    private String name;
+    private ResourceType type;  // Index [Search all resources of type xx]
+    private String feed;    // Index; But not sure if feeds are still in play if the inventory is built from directly prometheus scans
+    private String rootId;  // Nullable; Index [Search all resources under root xx]
+    private Map<String, String> properties;
+
+    // Maybe "children" and "metrics" should be removed from "Resource" and put in another class that represents the whole tree
+    private Collection<Resource> children;
```
lucasponce: Perhaps this can be modelled in a lazy way.
The design is good, but probably I will do this in two steps:
a) Give me Resource id1
b) Give me children of id1
So, I wonder if when we return a resource, we need to collect all children + metrics or that can be on demand.
(Or we can define this thinking on the flows we need to support).
Same with metrics, perhaps I want the metrics of a resource but I'm not interested in the children.

lucasponce: But yes, design is ok. Just thinking loud.
(My comments are not a formal revision, just catching up or making suggestions with all the feedback we are collecting)

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/oldmodel/Metric.java
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Metric describes a single metric that is sent out from a feed. Each metric has a unique ID and a type. Metrics live
+ * in an environment and can be "incorporated" by {@link Resource resources} (surprisingly, many resources can
+ * incorporate a single metric).
+ *
+ * @author Lukas Krejci
+ * @since 0.0.1
+ */
+@ApiModel(description = "A metric represents a monitored \"quality\". Its metric type specifies the unit in which" +
```
lucasponce: Side note.
In Hawkular Alerting we have migrated from Swagger annotations into Hawkular custom annotations which are more flexible for defining our api (and also it works better with vertx handlers).
It is in my todo to put them in commons but I see it can be useful for there too.

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Resource.java
jotak: +1 for lazy loading

lucasponce:
@jotak as this is target mwm branch I vote to merge this and discuss/refine on the go in a better agile approach.
In alerting 2.x we followed that strategy (to avoid long/large) PR discussions and the experience has been really good.
If you are agree I can merge and we can continue discussing on the code.

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Metric.java
```diff
@@ -16,175 +16,13 @@
...
-            public Blueprint build() {
-                return new Blueprint(metricTypeId, id, name, collectionInterval, properties, outgoing, incoming);
-            }
-        }
-    }
+public class Metric {
+    private String id;
+    private String name;
+    private String feed;    // Index; But not sure if feeds are still in play if the inventory is built from directly prometheus scans
+    private MetricType type;  // Index [Search all metrics of type xx]
```
jotak: @lucasponce here the ideal index would be "metric.type.id" (for query such as "search all metrics whose type id is xx")
Is it easy to do?

lucasponce: Yeah, I think that perhaps MetricType is stored separately from Metric, then, here we could have a typeId that serves as the link with the MetricType and also it is a field for indexing.

lucasponce:
As discussed with Joel, for this PoC/Preliminar steps we are going to work in a more "agile" approach.
So, merging and refining (unless some blocker found).

PR #116, opened 2017-09-12
Model proposal

jotak:
- Immutable model
- Model content initialized from full-args ctor
- Relational content is referred in two ways:
  - primarily with ids, stored in model
  - with lazy-loaded object reference


lucasponce:
+1

PR #117, opened 2017-09-13
Simplified model

jotak:
- Not directly tied to prometheus (prometheus metric ids should be stored in "properties" of the Metric objects
- Lazy-loaded stuff
- No "OperationType" anymore, assuming it's ok to merge it in "Operation"
- No "MetricType" anymore, assuming it's ok to merge it in "Metric"
- Remove old model


lucasponce:
+1

lucasponce:
Minor comment, not related to the model, to avoid extra messages we could de-activate travis on this branch or perhaps check the error in the CI if it is related to something touched during this effort.

jotak:
The error was about license not updated, should now be fixed, let's see in the next PR if there's still a problem...

PR #118, opened 2017-09-13
Remove unneeded maven modules, rename to hawkular-inventory-model

jotak:


jotak:
Error here is on command gateway because it needs some pieces of inventory that I removed

jotak:
Outdated

PR #119, opened 2017-09-22
Inventory updates

jotak:
- REST API: manage delete
- Serve agent config files
- Remove Metrics index in ISPN / not searchable by metric anymore (metrics are now directly attached to their resource)
- Service unit tests

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/pom.xml
+++ b/hawkular-inventory-parent/hawkular-inventory-service/pom.xml
@@ -236,6 +236,30 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <!-- FIXME: this is for arquillian only, there's probably another way to do it -->
+        <artifactId>maven-resources-plugin</artifactId>
+        <executions>
+          <execution>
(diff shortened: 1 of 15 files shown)


hawkular-inventory-parent/hawkular-inventory-service/pom.xml
```diff
@@ -236,6 +236,30 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <!-- FIXME: this is for arquillian only, there's probably another way to do it -->
```
jotak: @lucasponce / @jshaughn , I'm copying test resources in the WF deployment for integration tests, I don't know if there's another way to do that.
Basically what I want is to have my "test.yml" file under "$wf/standalone/configuration/hawkular" for tests only. But I don't want it to appear in the dist.

lucasponce: Yes, you need to add it in the @Deployment method of the test as a resource

lucasponce:
LGTM
Merging to consolidate.
We can refine later.

PR #123, opened 2017-09-25
Support for pagination and improvements in performance

lucasponce:
- Changed API to support pagination.
- Improvements on performance: tweak indexes and configuration.
- Additional tests to get measures on big scenarios.

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-model/src/main/java/org/hawkular/inventory/model/Resource.java
+++ b/hawkular-inventory-parent/hawkular-inventory-model/src/main/java/org/hawkular/inventory/model/Resource.java
@@ -26,7 +26,10 @@ import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import org.hibernate.search.annotations.Analyze;
+import org.hibernate.search.annotations.Field;
 import org.hibernate.search.annotations.Indexed;
+import org.hibernate.search.annotations.Store;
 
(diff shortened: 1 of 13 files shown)


hawkular-inventory-parent/hawkular-inventory-model/src/main/java/org/hawkular/inventory/model/Resource.java
```diff
@@ -44,9 +47,11 @@
     private final String name;
 
     @JsonInclude(Include.NON_NULL)
+    @Field(store = Store.YES, analyze = Analyze.NO, indexNullAs = Field.DEFAULT_NULL_TOKEN)
```
jotak: What's the options for "indexNullAs"? For information, "typeId" should never be null. "rootId" on the contrary can be null ; I don't know if it impacts anything here.

lucasponce: it has little impact, basically is to decide which token will be used to stored null values.
By default, null values are skipped into the indexing engine and we want them at least for rootId.
Today we don't procect/validate per field (which is not bad as main users is agent), so we could remove this validation on typeId in a future iteration when we sync use cases of the agent.

jotak:
LGTM - I imagine the indexed fields will need to be verified again later in the light of knowing the most common queries from MiQ

PR #124, opened 2017-09-25
start of jmx exporter configs

jmazzitelli:
I just want some place to put a jmx exporter config that I configured for our first attempt. This works on wildfly 10.x.

@jotak I found an issue we need to address. Since this just builds the war, the only place I can put it is in WEB-INF/classes (along with the ispn.xml file). However, we want the user to be able to create their own or override this default file. So we need to change the REST endpoints to first look in standalone/configuration as you have it now, however, if the file is not found, look for it in the WAR classloader (and thus will pick up any default files we ship with the war).

This will also help our current distribution model - which as it stands now is just a war and we ask devs to copy the war to any wildfly standalone/deployment folder. Without this new feature, we need to ask people to also copy some yaml files to standalone.configuration in addition to putting the war in the deployment folder.

This PR adds this new feature with tests.

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
@@ -17,6 +17,8 @@
 package org.hawkular.inventory.service;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -24,6 +26,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Scanner;
 import java.util.stream.Collectors;
 
 import javax.ejb.Local;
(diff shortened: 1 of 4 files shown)


lucasponce:
> So we need to change the REST endpoints to first look in standalone/configuration as you have it now, however, if the file is not found, look for it in the WAR classloader (and thus will pick up any default files we ship with the war).

+1 on this, It is not yet in place but also the ispn configuration should be also overwritten on disk.
A standalone/configuration/hawkular folder sounds good.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
```diff
@@ -210,8 +213,20 @@ public void deleteResourceType(String type) {
         try {
             byte[] encoded = Files.readAllBytes(configPath.resolve(fileName));
             return Optional.of(new String(encoded, "UTF-8"));
-        } catch (IOException e) {
-            return Optional.empty();
+        } catch (IOException ioe) {
+            try {
+                InputStream is = this.getClass().getClassLoader().getResourceAsStream(fileName);
+                String text = null;
+                try (Scanner scanner = new Scanner(is, StandardCharsets.UTF_8.name())) {
```
jotak: I guess the same way of reading should be used while reading at the initial location

PR #126, opened 2017-09-26
Inventory: add resource type to resource JSON in rest API

jotak:
- new class ResourceWithType
- update deserializer
- add index on ResourceType.id, because it's going to be a very common query
- update tests


jotak:
@lucasponce as you can see I added an index on ResourceType.id, I think it's necessary because it's going to be a very common query.

However I've found something surprising (at least to me), when there's a lot of Resource data, getting all resource types become very slow. Does it mean that entities are all merged in a single store?
This is a problem, getting all resource types should be super fast.
I've added a REST test to emphasize this: see `test019_shouldGetResourceTypes`, which is the same as `test005_shouldGetResourceTypes` but takes 6.5s versus 0.2s for the latter ; the only difference is the quantity of resources in store.

lucasponce:
Let me take a look, it seems that something is not indexing well.
I'm going to review your PR locally to find it out.

lucasponce:
At the end there is a known bug for this
https://issues.jboss.org/browse/ISPN-8344

So, I'm going to merge this one as we have commented, we can split the caches for resources and types separately.

PR #129, opened 2017-09-28
Update rest endpoints

jotak:
- Always use plural "resources"
- Use /resources/{id}/tree instead of /resources/tree/{id} (fixing potential conflict)
- Change "rootId" to "isRoot" in data model
- Merge endpoints that return Resources with different filters (get all root, get by type)


lucasponce:
> Change "rootId" to "isRoot" in data model

But then we will miss the child - parent relationship, isn't it ?

lucasponce:
LGTM

PR #132, opened 2017-09-28
Inventory: rename Operation.parameterTypes => parameters

jotak:


lucasponce:
it is not necessary to propragate this on tests data ?

jotak:
There's no test on operation parameters I think.. just operation names are checked. So thy're green.

PR #135, opened 2017-09-29
Remove Metric.collectionInterval

jotak:


jotak:
@jmazzitelli do we agree on that ? (removing collectionInterval)
Btw there may be other things to do. My understand is that "name" will hold the display name for MiQ (but not sure if it's needed), and we could put p8s info in the properties map. Or you can suggest other way.

jmazzitelli:
I think we are going to need more here to support Prometheus metrics. At minimum we'll need something additional. I can think of two possible ways we can go:

1) Have two fields "String metricFamilyName" and "List<String,String> labels"

or

2) Have one "String metricId" field that will be a full Prometheus metric identifier like "my_metric_seconds{label1="foo", label2="bar"}"

I don't know which of the two will be easiest for use to use and clients to consume. But I do know we'll eventually need either 1) or 2) above.

jshaughn:
To keep going on the discussion here.   From what I understand, we will be able to set the metric "job" label to the feedId (this happens server-side, I believe, at the same time the instance tag is applied).  If we can add an "resourceId" label to each metric definition in the jmx-exporter config, it will be a pretty solid tie between metric and resource, giving us the (feedId,resourceId) tuple.

PR #139, opened 2017-10-02
Replace children with parent in Resource model

jotak:
I listed some pros & cons about using parent versus children in model:

PRO:
- Easier update from the agent: when a resource is added or deleted, no need to search for its parent to update it too. Correctness of the tree not threatened.
- No need for "isRoot" field & index anymore.

CONS:
- Must be indexed to satisfy queries such as: "find all resources whose parent is xxx", typically to build subtree.
- Cannot get children ids from raw resource anymore. So they will not be visible either when getting a resource not in "tree" mode.


lucasponce:
I see the pros/cons.
Performance speaking, no querying but using the cache key is fastest way and impact is low.
Indexing adds some overhead, but it's a tradeoff.
It is in my todo to extend the perf use cases, if we start to visualizing the flows to add or query we can test if we are good for the threshold discussed (to be able to manage thousands of EAP per node).

lucasponce:
LGTM

PR #141, opened 2017-10-03
Use filter object to get resources

jotak:


hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/ResourceFilter.java
```diff
@@ -0,0 +1,71 @@
...
+        this.rootOnly = rootOnly;
+        this.feedId = feedId;
+        this.typeId = typeId;
+    }
+
+    public static ResourceFilter rootOnly() {
+        return new ResourceFilter(true, null, null);
+    }
+
+    public static ResourceFilter ofType(String typeId) {
```
lucasponce: I don't have any strong opinion, but as @jmazzitelli started to introduce builders, why don't maintain same semantic ?
I see this one has more "lambda" style, and I'm fine just wondering if we want to maintain styles or this is ok.
LGTM in any case.

PR #146, opened 2017-10-04
Add export endpoint + split tests

jotak:
In the same time, refactor tests to separate perf & rest tests so that "export" is not run after perf tests.
New common file for test resources definitions

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/InventoryService.java
+++ b/hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/InventoryService.java
@@ -16,6 +16,8 @@
  */
 package org.hawkular.inventory.api;
 
+import java.io.IOException;
+import java.io.OutputStream;
 import java.util.Collection;
 import java.util.Optional;
 
@@ -120,5 +122,5 @@ public interface InventoryService {
      */
     boolean isRunning();
 
-    Import buildExport();
+    void buildExport(OutputStream os) throws IOException;
 }
--- a/hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/InventoryService.java
+++ b/hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/InventoryService.java
@@ -119,4 +119,6 @@ public interface InventoryService {
      * @return true is InventoryService is Up and Running
      */
     boolean isRunning();
+
+    Import buildExport();
 }
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
@@ -61,6 +61,17 @@ public class InventoryHandlers {
      */
 
     @GET
+    @Path("/export")
+    @Produces(APPLICATION_JSON)
+    public Response exportInventory() {
+        try {
+            return ResponseUtil.ok(inventoryService.buildExport());
+        } catch (Exception e) {
+            return ResponseUtil.internalError(e);
+        }
+    }
+
+    @GET
     @Path("/get-inventory-config/{templateName}")
     @Produces(TEXT_PLAIN)
     public Response getAgentConfig(@PathParam("templateName") final String templateName) {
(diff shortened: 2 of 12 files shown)


hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/InventoryService.java
```diff
@@ -119,4 +119,6 @@
      * @return true is InventoryService is Up and Running
      */
     boolean isRunning();
+
+    Import buildExport();
```
lucasponce: Naming types of variables is very a matter of taste.
In Alerting we finished up using plural, Import object perhaps now is weird, "Resources", "Inventory" or something that points to all the data, can be good.

lucasponce: Also, I think this should be paginated.
If not a OOME is on the way.

jotak: I was looking for nice names, +1 for Inventory

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
```diff
@@ -61,6 +61,17 @@
      */
 
     @GET
+    @Path("/export")
+    @Produces(APPLICATION_JSON)
+    public Response exportInventory() {
```
lucasponce: Same here, pagination should be needed.
If we don't want pagination we can treated it as multipart streaming, and pagination is only in the InventoryService.
I don't know, but this is hit a OOME in the preliminar tests of QE.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
```diff
@@ -291,4 +284,53 @@ private boolean isEmpty(Collection c) {
         return c == null || c.isEmpty();
     }
 
+    @Override
+    public void buildExport(OutputStream os) throws IOException {
+        try (Writer writer = new BufferedWriter(new OutputStreamWriter(os))) {
+            writer.write("{\"types\":[\n");
```
jotak: @lucasponce do you see a better way to do it, than string-writing json? I'm stuck with StreamingOutput + paginated combination, I cannot jsonify with Jackson a yet-incomplete object. You think it's ok?

lucasponce: I know that jackson offers some streaming cappabilities, but honestly I haven't tested them yet, so only I read API and examples. Perhaps is worth to take a look and investigate this together.
I remember that in alerting the use was different as the format was not a json per se, but chunks of jsons, but on this case it seems we could rely on a library as the output is a fully json object.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
```diff
@@ -279,4 +284,53 @@ private boolean isEmpty(Collection c) {
...
+                    } else {
+                        first = false;
+                    }
+                    writer.write(JsonUtil.toJson(r) + "\n");
+                }
+                hasMore = batch.size() == MAX_RESULTS;
+                offset += MAX_RESULTS;
+            }
+            writer.write("]}\n");
+            writer.flush();
```
lucasponce: I think here the response is flushed at the end, I think it should be flushed also in intermediate steps too.
As commented, I would need to run some tests to see if I am right or I miss something in my comment.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
lucasponce: I remember I ran this example locally
http://www.baeldung.com/jackson-streaming-api
but I didn't investigate with detail if I can flush in the intermediate, to check that definetly we can control the buffer and avoid to have all data in memory.

jotak: Thanks, I've already seen it but it went out of my head :)

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
jotak: I've updated with JsonGenerator, adding intermediate flush, the tests are still OK

lucasponce:
Thx @jotak LGTM

PR #147, opened 2017-10-04
Add Service to just get direct child resources of a parent.

jshaughn:
I know there is already getTree but this helps the UI expand a tree incrementally, and I think it is generally useful to retrieve only immediate children.

@jotak Please review, thanks.


lucasponce:
+1

jotak:
LGTM

PR #148, opened 2017-10-06
Model: merge "Resource" and "ResourceWithType"

jotak:
I changed the model as it was suggested by chat. However I'm not keen on that model, but leave this PR for debate.

I think merging "Resource" and "ResourceWithType" only gives the illusion of a simpler model, but now objects of type Resource can have various states (depending on lazy-loaded things) and "getType" with no argument could sometimes give null, sometimes a ResourceType, depending on the context.

To me, this is not clean: I prefer when context is clearly enforced by type (so with a separate class), it reduces mistakes and the API consumer can without doubt know what it can do with it.

What do you folk think?


jotak:
PS: my initial proposal, if we keep separate classes, was to rename "Resource" to "RawResource" and "ResourceWithType" to "Resource".

lucasponce:
From the Java perspective I can see a long debate as I think any solution can be good or bad.
Basically we are discussing if resource might have populated one or other field depending on the method which is invoked like:

Resource r1 = getFetchA()
Resource r2 = getFetchB()

Yes, here there could be a debate that ResourceA should be a different type and ResourceB to differentiate this.

But, having in mind the REST layer, which serialize in a JSON, I don't see too much difference, as the user needs to know that one method will have one field or another.

Anyway, the Java API is purely internal, only used by agent and cmdgw, and the rest api will be used by the ruby client, were JSON response should be described in any case.

Personally I like the join approach.

So, I think this impacts more on the Java API consumers than the ruby client REST API.

jotak:
How about the doc, if at some point we decide to properly document the REST API?

Both Resource and ResourceWithType are exposed in the API, the former for import/export endpoints, the latter for all query endpoints. So the design choice would leak into the doc. Moreover if we use auto-generated doc, that would produce some garbage with unused fields depending on context.

lucasponce:
> How about the doc, if at some point we decide to properly document the REST API?

In alerting 2.x we proposed a Hawkular annotations as Swagger is limited to document REST endpoint, specially in types, so we proposed to document that properly, as an user of the REST API is not going to get any class or Java object, but a JSON object.

But anyway, we can give arguments in one way or another, whoever has the ownership to decide, as commented, I am not convinced that one or another is good / bad, better than the other.

It is a matter of taste and feel comfortable, I propose that you decide here.

jshaughn:
Personally, I like the joined approach that limits to a single type, even if that type may not be fully populated.  That is not unlike many APIs that deal with lazy-loaded entities, like Hibernate and such.  A little documentation should be able to set expectations.

jotak:
Ok, then, majority wins :)
Another thing I'd like to do while cleaning up the api, is to merge h*-inventory-model inside a "model" package in h*-inventory-api. I think these two are anyway strongly coupled and should come together. Do you see something wrong with that?

lucasponce:
> Another thing I'd like to do while cleaning up the api, is to merge h*-inventory-model inside a "model" package in h*-inventory-api. I think these two are anyway strongly coupled and should come together. Do you see something wrong with that?

+1

In a future step, we would want to have an internal model to not fetch the hibernate search annotations with the api, but definitely this is a future enhancement, no for now (just comment here more for myself to not forget thinking about it)

jotak:
Hmm actually... I'm sorry but I insist. I'm currently dealing with integration test issues consecutive to lazy loading. It's not that they can't be fixed, sure they can, but the uncertainty of having lazy-loaded things in memory makes things more prone to side-effects.

The example I'm hitting right now is in the "export" test, the /export endpoint should normally send resources without their loaded children, but since it's ISPN, things are in memory, a previous call of whatever query did load them, so they're in the returned json.

That's exactly the point of having an immutable model, but here we have a just partially immutable model and the side-effects start to appear.

jmazzitelli:
@jotak do what you think is best.

lucasponce:
> The example I'm hitting right now is in the "export" test, the /export endpoint should normally send resources without their loaded children, but since it's ISPN, things are in memory, a previous call of whatever query did load them, so they're in the returned json.

I would suggest to take a look into the alerting 2.x branch.

We had an internal model prefixed with Ispn* for this case, where what is stored is  inmutable.

lucasponce:
Further discussion of this topic on #154 I think this can be closed, otherwise, please @jotak reopen it.
Thx.

PR #149, opened 2017-10-07
Inmutable objects in backend and fixes in performance tests

lucasponce:
@jotak, this is related our discussion on friday.
This approach was followed in Alerting 2.x, basically the idea is to separate the object used to store in the backend vs the object used in the Java API, and make it inmutable.
Also, this PR fixes performance tests and makes them mandatory in the profile.

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
@@ -45,6 +45,8 @@ import org.hawkular.inventory.log.InventoryLoggers;
 import org.hawkular.inventory.log.MsgLogger;
 import org.hawkular.inventory.model.Resource;
 import org.hawkular.inventory.model.ResourceType;
+import org.hawkular.inventory.service.ispn.IspnResource;
+import org.hawkular.inventory.service.ispn.IspnResourceType;
 import org.infinispan.Cache;
 import org.infinispan.query.Search;
 import org.infinispan.query.dsl.FilterConditionContextQueryBuilder;
@@ -106,9 +108,9 @@ public class InventoryServiceIspn implements InventoryService {
         if (isEmpty(resources)) {
             return;
         }
-        Map<String, Resource> map = resources.stream()
+        Map<String, IspnResource> map = resources.stream()
                 .parallel()
-                .collect(Collectors.toMap(r -> r.getId(), r -> r));
+                .collect(Collectors.toMap(r -> r.getId(), r -> new IspnResource(r)));
         resource.putAll(map);
     }
 
@@ -122,9 +124,9 @@ public class InventoryServiceIspn implements InventoryService {
         if (isEmpty(resourceTypes)) {
             return;
         }
-        Map<String, ResourceType> map = resourceTypes.stream()
+        Map<String, IspnResourceType> map = resourceTypes.stream()
                 .parallel()
-                .collect(Collectors.toMap(rt -> rt.getId(), rt -> rt));
+                .collect(Collectors.toMap(rt -> rt.getId(), rt -> new IspnResourceType(rt)));
         resourceType.putAll(map);
     }
 
@@ -149,7 +151,7 @@ public class InventoryServiceIspn implements InventoryService {
         if (isEmpty(id)) {
             throw new IllegalArgumentException("Resource id must be not null");
         }
-        return Optional.ofNullable((Resource) resource.get(id));
+        return Optional.ofNullable((IspnResource) resource.get(id)).map(IspnResource::getResource);
     }
 
     @Override
@@ -165,7 +167,7 @@ public class InventoryServiceIspn implements InventoryService {
 
     @Override
     public ResultSet<ResourceWithType> getResources(ResourceFilter filter, long startOffset, int maxResults) {
-        QueryBuilder qb = qResource.from(Resource.class);
+        QueryBuilder qb = qResource.from(IspnResource.class);
         FilterConditionContextQueryBuilder fccqb = null;
         if (filter.isRootOnly()) {
             fccqb = qb.having("parentId").isNull();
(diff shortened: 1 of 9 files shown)


hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/ispn/IspnResource.java
```diff
@@ -0,0 +1,97 @@
...
+        return typeId;
+    }
+
+    public String getParentId() {
+        return parentId;
+    }
+
+    public Resource getResource() {
+        if (resource == null) {
+            return resource;
```
jmazzitelli: why 'return resource'? Shouldn't this be return null (yes, I realize resource is null because it wouldn't get here unless "if resource == null" is true, but still, seem odd and confusion). Should be "return null" just to be clear what it is that you intend to happen here (that is, return null when resource isn't set).

lucasponce: Fixed.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
```diff
@@ -149,7 +151,8 @@ public void deleteResourceType(String typeId) {
         if (isEmpty(id)) {
             throw new IllegalArgumentException("Resource id must be not null");
         }
-        return Optional.ofNullable((Resource) resource.get(id));
+        IspnResource ispnResource = (IspnResource) resource.get(id);
+        return Optional.ofNullable(ispnResource == null ? null : ispnResource.getResource());
```
jotak: not important, but that could be "return Optional.ofNullable(ispnResource).map(IspnResource::getResource)"

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
```diff
@@ -209,7 +218,8 @@ public void deleteResourceType(String typeId) {
         if (isEmpty(typeId)) {
             throw new IllegalArgumentException("ResourceType id must be not null");
         }
-        return Optional.ofNullable((ResourceType) resourceType.get(typeId));
+        IspnResourceType ispnResourceType = (IspnResourceType) resourceType.get(typeId);
+        return Optional.ofNullable(ispnResourceType == null ? null : ispnResourceType.getResourceType());
```
jotak: same here

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
```diff
@@ -279,17 +291,20 @@ public boolean isRunning() {
...
+                .map(r -> ((IspnResource) r).getResource())
+                .collect(Collectors.toList());
     }
 
     private ResourceType getNullableResourceType(String typeId) {
         if (isEmpty(typeId)) {
             return null;
         }
-        return (ResourceType) resourceType.get(typeId);
+        return ((IspnResourceType) resourceType.get(typeId)).getResourceType();
```
jotak: return value here can be null if the type doesn't exist, so it's missing a null-check

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/ispn/IspnResource.java
```diff
@@ -0,0 +1,97 @@
...
+    @Field(store = Store.YES, analyze = Analyze.NO, indexNullAs = Field.DEFAULT_NULL_TOKEN)
+    private final String feedId;
+
+    @Field(store = Store.YES, analyze = Analyze.NO, indexNullAs = Field.DEFAULT_NULL_TOKEN)
+    private final String typeId;
+
+    @Field(store = Store.YES, analyze = Analyze.NO, indexNullAs = Field.DEFAULT_NULL_TOKEN)
+    private final String parentId;
+
+    private final Resource resource;
```
jotak: I'm not sure if that's ok. If the whole stateful "Resource" object is serialized with IspnResource, then we may have the same side-effects that I hit before, with lazy-loaded "children" being serialized as well here? The mutability of "resource" is going to leak to IspnResource.

jotak: Oh, no because it's cloned, I understand.
But i'm still in favor of Immutable "Resource" :)

jotak: Thinking again about it... it's cloned at construction, but then it's provided through getter without re-cloning, so it can still be mutated. I think it's a problem.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
lucasponce: +1

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
lucasponce: +1

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
lucasponce: +1 good catch, I missed this one

lucasponce:
I have addressed the comments of the PR.
I think this approach is interesting, as it separates API vs backend objects, and not bringing the hibernate-search dependencies to the Java API users (mostly cmdgw and listeners, as agent will serialize).
About the discussion Resource mutable vs several Resource classes inmutable depending if a resource brings children or type, is still being discussed in the #148 I guess this one could get in, and in the previous PR we can still discussing.

jotak:
Agree, I didn't mean to block this PR

PR #150, opened 2017-10-09
Bring back UpdateCollectionIntervals commands

lucasponce:
In the context of Prometheus, we dropped collection interval commands.
Restoring these classes to enable CmdGw sync correctly with the agent.


lucasponce:
Please @jotak could you take a look (if you can).

jotak:
+1, but does it mean we must also take it back in Metric model? The "collectionInterval" field was removed.

PR #152, opened 2017-10-10
Add delete all endpoints

jotak:
so that itests run better
Also add "@get /types/{typeId}" endpoint that was curiously missing


lucasponce:
LGTM
I wonder if we can use this PR to add a delete many (resources and types), passing a list of ids (resources and types).
I guess that will be useful too, and it could be a low hanging fruit.

lucasponce:
LGTM

PR #153, opened 2017-10-10
Add config map in resources

jotak:
Also refactor resource class constructors so that they're consistent one another


jotak:
@jmazzitelli here's the config map

PR #154, opened 2017-10-11
Full-immutable model

jotak:
- Rename Resource => RawResource
- Rename ResourceWithType => Resource
- Remove lazy-loaded (thus mutable) fields in model
- Remove now unnecessary clone methods in Ispn* model


lucasponce:
Today we have ResourceNode, Resource and RawResource.
Personally, I would like to unify them but I see that avoid cloning will save JVM pressure.
So, I won't dig more +1 from my side.
Now is time from @jshaughn as he also participated on the previous discussion.

jshaughn:
+1, I like the naming better but I guess we still have two classes in play (Resource and RawResource).  I'd like to see this unified/simplified to just Resource where we'd persist/index the "raw" version.  I' not against transient fields, like resourceType, that get optionally populated.  I'm not convinced the cloning is a blocking issue to the design.  Anyway, this can be merged as not to slow things down but we should should consider unification and if we are going to do it, do it sooner than later.

jotak:
Ok, so I'll merge it and rewrite my arguments too, for the record:

- Having several classes is strong&safe typing and I definitely believe in safe typing. Thanks to it, any user of that class knows exactly what it can or cannot do just because of its type, and I think this is good.
On the other hand, having a single, multi-shaped class *looks* simpler but I don't think it *is* simpler, because its usage is less obvious (will I have data in that field? What does it mean if I haven't?) - and in api design I think usage takes priority over the rest, even if, for the moment, the usage is restricted to some of our own components.

By the way, if the number of duplicated lines of code is a problem, we can still abstract them (I'm fine with, or without it) - that would just make one more class, not sure you'd like. But duplication is not always bad.

- Immutable objects have proved to be useful & safer for many reasons (like parallelism or being side-effect free), I won't be listing all blog posts on that subject, I know our use case is limited enough to easily keep that under control, but why not adopting good practice from the beginning?

- Like said, the other argument is not about api design but performance, I'm not sure how big is the impact, but with mutable model all ispn in/out access would necessitate a clone operation.

lucasponce:
@jotak for the record, too.
I see the advantage to avoid unnecessary cloning, but others arguments are arguable in the meaning that with one option or another, probably the model will be correct.

PR #155, opened 2017-10-11
Reindex caches at startup

lucasponce:


lucasponce:
@jotak, as commented, re-index at startup (default true, but possibility to disable via configuration).
And minor perf tweaks.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryConfig.java
```diff
@@ -91,6 +97,17 @@ public void init() {
                 log.errorInventoryCacheNotFound();
                 throw new IllegalStateException("Inventory query factory for resource_type cache is not found");
             }
+            if (ispnReindex) {
+                log.infoStartInventoryReindex();
+                long startReindex = System.currentTimeMillis();
+                SearchManager searchResourceManager = Search.getSearchManager(resource);
+                searchResourceManager.getMassIndexer().start();
```
jotak: If it's a time-consuming operation, there may be a benefit to run them in parallel ?

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
```diff
@@ -111,7 +112,7 @@ public void addResource(Collection<Resource> resources) {
         Map<String, IspnResource> map = resources.stream()
                 .parallel()
                 .collect(Collectors.toMap(r -> r.getId(), r -> new IspnResource(r)));
-        resource.putAll(map);
+        resource.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES).putAll(map);
```
jotak: out of curiosity, is it for better performances?

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryConfig.java
lucasponce: Mmm, yes, you are right, as the caches are splitted it might help.
I will add that.

lucasponce: As those are separated caches it should, +1, I will add this suggestion.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryServiceIspn.java
lucasponce: Yes, put and remove operations than don't use the returned value can get benefit for this.
To be honest, in local modes as we have today the gain is not high, but definitely will help in clustering modes.
I introduced here as sooner or later we will work on horizontal scalability.

jotak:
LGTM
Ready for merge?

PR #157, opened 2017-10-13
HAWKULAR-1264 Rest API self discovery

jotak:
Add resteasy dependency (provided) for REST endpoints introspection

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
@@ -21,7 +21,6 @@ import static javax.ws.rs.core.MediaType.TEXT_HTML;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
 
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -58,8 +57,9 @@ import org.hawkular.inventory.handlers.ResponseUtil.ApiError;
 import org.hawkular.inventory.log.InventoryLoggers;
 import org.hawkular.inventory.log.MsgLogger;
 import org.jboss.resteasy.core.Dispatcher;
-import org.jboss.resteasy.core.ResourceMethodInvoker;
 import org.jboss.resteasy.core.ResourceMethodRegistry;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * @author Jay Shaughnessy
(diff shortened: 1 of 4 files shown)


jotak:
See the result here: http://i.imgur.com/mWjMzLA.png

Please, comment if you'd like something more shiny or different...

jotak:
Json version:
https://paste.fedoraproject.org/paste/hYg55sx5uiVp3LRHX5gLCQ

jotak:
(from my side, can be merged)

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
```diff
@@ -53,8 +58,9 @@
 public class InventoryHandlers {
 
     private static final MsgLogger log = InventoryLoggers.getLogger(InventoryHandlers.class);
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
lucasponce: Suggestion (minor), JsonUtil.getMapper() has also same object and it is used in other handlers..

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
```diff
@@ -64,6 +70,45 @@
      */
 
     @GET
+    @Path("/")
```
lucasponce: To choose between this json and html output, client should adjust the "accept" header, I think.
If there is anything else it could be nice to have a comment here, I will add it into the rest documentation later.

jotak: Yes, it's how it works. However there's warning messages in the logs that are bugging me :(
We have the same in hawkular metrics. IIRC it says something like "Two methods found for endpoint, selecting one". Which seems stupid because there's the content negotiation. I don't know if we can get rid of them (without changing the endpoint)

lucasponce: Yes, I think is something in the RESTEasy engine.
To get rid a single endpoint can be used, adjusting the output manually checking the accept header.
Some extra code but that would get rid of warnings in log, I think.

pilhuhn: We should ask the RE folks first. I don't think doing manual parsing is good here, as the Accept header could contain more complex expressions and with weights like application/json;w=0.3,text/html;w=0.7 (not the exact syntax), where despite a/j being listed 1st in reality text/html is preferred.

jotak:
I'm reading about / trying the “quality from server” factor in jax-rs. It may fix the problem.

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/handlers/InventoryHandlers.java
jotak: I'm reading about / trying the “quality from server” factor in jax-rs. It may fix the problem.

jotak:
IMO it could be merged. The log.warn issue could be tracked in a separate ticket.

jotak:
for the record: https://issues.jboss.org/projects/HWKINVENT/issues/HWKINVENT-210

PR #158, opened 2017-10-16
HAWKULAR-1263 Document the Inventory REST API

lucasponce:
- Move to commons lightweight documentation annotations used in Alerting 2.x
- Adjust groovy script to support generic type use cases
- Prepare travis scripts to upload documentation into github

Note that this PR won't publish documentation into hawkular.github.io yet for this it should be decided when to upload doc (on every commit, on releases only, from one branch or another branch, etc).

Once this is decided, secure settings should be updated into travis.yml but all scripts should be working.


jshaughn:
LGTM, thanks @lucasponce .

PR #160, opened 2017-10-16
Hawkular-1279 Basic Authorization for Inventory

jshaughn:
Do Not Merge until clients (agent, ruby, ui) are prepared to send the Authorization header.   I believe the agent is already doing this.  Need to check ruby client and add something to the ui.

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/main/webapp/WEB-INF/web.xml
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/main/webapp/WEB-INF/web.xml
@@ -24,4 +24,33 @@
 
   <display-name>Hawkular Inventory Service</display-name>
 
+  <security-constraint>
+    <web-resource-collection>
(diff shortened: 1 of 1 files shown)


jshaughn:
Because browsers (Chrome, FF tested) detect authorization failures, prompt for user/pass, and then add the header automatically on subsequent requests, the UI does not need any immediate changes.  We could optionally add the patternfly login screen going forward.

jshaughn:
@josejulio indicates that the ruby client is sending the Authorization header already, as before.  So I think probably the only issue is ensuring the itests pass the header.  I can add that tomorrow...

lucasponce:
I have one question. Why to place here the security details and not in hServices overwritten the web.xml / jboss-web.xml descriptors ?
I think security should be defined in a single place and maven offers tool to overwrite this config when assembling the final artifacts in hServices.

jshaughn:
I don't really have anything for or against locating security with the component, I belive it has been done this way in the past.  Because you have a preference I will see about doing this in services.

PR #161, opened 2017-10-17
HAWKULAR-1273 Inventory Monitoring

lucasponce:
- Memory, disk, cpu and inventory stats

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/test/java/org/hawkular/inventory/service/InventoryServiceIspnTest.java
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/test/java/org/hawkular/inventory/service/InventoryServiceIspnTest.java
@@ -20,6 +20,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.io.ByteArrayOutputStream;
+import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -31,12 +32,14 @@ import java.util.Optional;
 import org.hawkular.inventory.Resources;
 import org.hawkular.inventory.api.ResourceFilter;
 import org.hawkular.inventory.api.model.Inventory;
+import org.hawkular.inventory.api.model.InventoryHealth;
 import org.hawkular.inventory.api.model.RawResource;
 import org.hawkular.inventory.api.model.Resource;
 import org.hawkular.inventory.api.model.ResourceNode;
 import org.hawkular.inventory.api.model.ResourceType;
 import org.hawkular.inventory.api.model.ResultSet;
 import org.infinispan.Cache;
+import org.infinispan.configuration.cache.SingleFileStoreConfiguration;
 import org.infinispan.manager.DefaultCacheManager;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.junit.Before;
@@ -53,6 +56,7 @@ public class InventoryServiceIspnTest {
     private static final String ISPN_CONFIG_LOCAL = "/hawkular-inventory-ispn-test.xml";
     private static EmbeddedCacheManager CACHE_MANAGER;
     private final InventoryServiceIspn service;
+    private final InventoryStats inventoryStats;
 
     static {
         try {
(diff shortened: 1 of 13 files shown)


mwringe:
Is this something we will eventually be able to pull into mwp with all of the other metrics we can collect from wildfly instances?

jmazzitelli:
This isn't a good implementation for us because our agent won't actually be able to collect these metrics itself (looks like this just exposes metrics over some custom REST endpoint).

Instead, you should expose the metrics via JMX so we can use the Jolokia stuff. Or, use the Prometheus client to expose these metrics as a Prometheus endpoint (their Java library is very easy to use).

Doing it this way really won't allow us to "eat our own dog food" - the agent won't be able to collect these metrics.

jmazzitelli:
Here's an example using a servlet to expose the Prometheus metrics endpoint: 

https://github.com/prometheus/client_java/blob/master/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java

Here's the README to show the API:

https://github.com/prometheus/client_java/blob/master/README.md

This even gives you the internal MemoryMXBean and other internal JMX mbeans for free - so you'll get memory, CPU, etc. metrics with a single line. From the README:

"The Java client includes collectors for garbage collection, memory pools, JMX, classloading, and thread counts. These can be added individually or just use the DefaultExports to conveniently register them."

hawkular-inventory-parent/hawkular-inventory-service/src/test/java/org/hawkular/inventory/service/InventoryServiceIspnTest.java
```diff
@@ -323,5 +332,11 @@ public void shouldGetLargeExport() throws IOException {
                 .hasSize(maxTypes + 3)
                 .extracting(ResourceType::getId)
                 .contains("EAP", "RT0", "RT100", "RT199");
+
+        InventoryHealth health = service.getHealthStatus();
+        assertThat(health.getInventoryStats()).isNotNull();
+        assertThat(health.getInventoryStats().getNumberOfResources() > 10000);
```
jotak: Here, no actual assertion is made... that would be `assertThat(health.getInventoryStats().getNumberOfResources()).isGreaterThan(10000)` , and similar below.

PS: I introduced assertj during the POC as I think it's an awesome lib, especially when dealing with collections, but I hope it doesn't cause frustration... If so let's talk about it

jmazzitelli:
The alternative is to simply convert your InventoryHealth object to an MBean (should be very simple to do with not much changes). You then just register your InventoryHealth MBean to the platform MBeanServer and we can then use JMX for the agent to collect the data.

jotak:
Mazz, it could be that both approaches are correct and complementary. The implementation here looks like an health check implementation like the one described here (by @pilhuhn not later than today!) : https://developers.redhat.com/blog/2017/10/17/monitoring-aspects-eclipse-microprofile-1-2/

jmazzitelli:
@jotak Yes, I think we can do both. As I say, it should be pretty easy to convert that InventoryHealth object to a JMX MBean, and just register it on the platform MBeanServer. You can leave everything else as is and just expose that InventoryHealth data via REST as well. But we should make that object an MBean to also show it via JMX.

lucasponce:
> Is this something we will eventually be able to pull into mwp with all of the other metrics we can collect from wildfly instances?

It is not designed in that way, the JIRA specifically defined an endpoint in inventory to serve this data.

jmazzitelli:
OK, after seeing what @pilhuhn did - you might be able to use this as-is - looks like Prometheus can collect metrics from REST endpoints. See: 

https://twitter.com/pilhuhn/status/918890934116470784

jmazzitelli:
Of course, from a Hawkular/MWM perspective, it means we now must support REST endpoints for metrics, not just JMX/Jolokia and normal Prometheus endpoints.

lucasponce:
> This isn't a good implementation for us because our agent won't actually be able to collect these metrics itself (looks like this just exposes metrics over some custom REST endpoint).

> Instead, you should expose the metrics via JMX so we can use the Jolokia stuff. Or, use the Prometheus client to expose these metrics as a Prometheus endpoint (their Java library is very easy to use).

> Doing it this way really won't allow us to "eat our own dog food" - the agent won't be able to collect these metrics.

I see the arguments. Basically we are discussing if inventory should have a standard MBean and let others collect this for integration with Prometheus or others, or just to add an self-health endpoint with this information to not required additional info.

Also, an alternative could be to create a MBean, and let the MBean be readed from anyplace but also offers the endpoint for this use case.

hawkular-inventory-parent/hawkular-inventory-service/src/test/java/org/hawkular/inventory/service/InventoryServiceIspnTest.java
lucasponce: Ok, I will re-arrange this to the assertj syntax.

lucasponce:
Ok, after the discussion I will add an MBean with this Inventory metrics and maintain the /health endpoint too.

lucasponce:
Update this PR to expose a MBean with Inventory metrics (metrics for memory and cpu are already exposed in other standard MBeans so no need to duplicate this info in the MBean).

lucasponce:
Found a couple of things I want to change.
Please, don't merge yet.

jmazzitelli:
LGTM

lucasponce:
As metric collection from ISPN can be a expensive task, the design is changed to collect info asynchronously.
Then, we will hold the last health information and that will be the information supplied to the user.
(Basically, ISPN uses a similar way).

lucasponce:
After discussion, agreement that fetching metrics on demand.

jmazzitelli:
LGTM

PR #169, opened 2017-10-25
HAWKULAR-1285 Add support for ISPN 8.5

lucasponce:
- Enable an Inventory built based on ISPN 8.5
- Add itest support for EAP+JDG combination

This PR split profiles into hawkular-inventory-service and hawkular-inventory-itest to support ISPN and JDG.

The upstream version is packaged in the inventory .war archive and tested on a Wildfly Server, meanwhile JDG is deployed as modules in an EAP, therefore inventory.war should be package differently depending of the target.

The commands 

```
mvn clean install
```
and

```
mvn clean install -Pitest
```
still work with ISPN and Wildfly and are managed by Travis CI as usual.

To test EAP and JDG, it is necessary to manually download and install an EAP + JDG library, and configure the JDG maven repository locally.

A EAP + JDG distribution can be prepared using a helper script like this:

```
EAP_ZIP=/home/lponce/Downloads/jboss-eap-7.0.8-full-build.zip
JDG_MODULES_ZIP=/home/lponce/Downloads/jboss-datagrid-7.2.0.DR2-eap-modules-library.zip

TARGET=/home/lponce/temp/inventory-perf

# Delete previous EAP and JDG installation
rm -Rf $TARGET/jboss-eap-7.0
rm -Rf $TARGET/jboss-datagrid-7.2.0-eap-modules-library

# Unzip EAP and JDG distributions
unzip $EAP_ZIP -d $TARGET
unzip $JDG_MODULES_ZIP -d $TARGET

# Deploy JDG library on EAP
cp -R $TARGET/jboss-datagrid-7.2.0-eap-modules-library/modules/* $TARGET/jboss-eap-7.0/modules
```

Then an JDG can be used with the commands

```
mvn clean install -Deap.home=/tmp/jboss-eap-7.0
```
and this one for running the full itest

```
mvn clean install -Pitest-eap -Deap.home=/tmp/jboss-eap-7.0
```

After some testing, all tests for inventory (unit tests, integration tests and performance tests) are running ok for ISPN 9.1.1 and JDG 7.2 (ISPN 8.5).


lucasponce:
> I'm just wondering what is the easiest way to handle several configs through several profiles: via file renaming in maven jobs as you did, or via a property that would give the ISPN file name. I don't have the answer, just asking the question :)

> If you think renaming is the best approach then LGTM

Yes, I know what you mean, I thought both options. I guess renaming was more natural and clear to me.
In this case, in runtime only one configuration will be needed, and also, as config content may vary, having one file per config helped me to see the steps clear.

jshaughn:
Already merged but the renaming approach seems OK to me as well.  Just wondering if we want specific versions in the file names or whether suing something like "upstream"/"downstream" would be easier as versions may change going forward?

lucasponce:
> Already merged but the renaming approach seems OK to me as well. Just wondering if we want specific versions in the file names or whether suing something like "upstream"/"downstream" would be easier as versions may change going forward?

My first draft included that, but as jdg71 and jdg72 are pretty different with its own configuration and special use cases, I think an splicit renaming clarified to me dealing with several configs.

PR #170, opened 2017-10-25
Constrain mbeans to whitelist to avoid grabbing all WF mbeans

jshaughn:
which is ineffecient and also generate errors.


hawkular-inventory-parent/hawkular-inventory-service/src/main/resources/wildfly-10-jmx-exporter.yml
```diff
@@ -19,11 +19,33 @@
...
+  "jboss.as:subsystem=datasources,xa-data-source=*,statistics=*",
+  "jboss.as:subsystem=transactions",
+  "jboss.as:subsystem=undertow,server=*,http-listener=*",
+  "jboss.as:subsystem=undertow,server=*,https-listener=*"
+]
 rules:
 
   # CONFIG
 
-  - pattern: "^jboss.as<path=(.+)><>path:(.+)"
```
jmazzitelli: I believe we want to keep the ^ in all these patterns because it will be more efficient. Pattern matcher will fail fast if the head of the string doesn't match - it doesn't have to search for a match through the string.

jshaughn: I think it's already built into the pattern here:

https://github.com/prometheus/jmx_exporter/blob/master/collector/src/main/java/io/prometheus/jmx/JmxCollector.java#L170

But I put them back to make you happy.

jmazzitelli: oh yeah! so it is. you didn't have to put it back :)

jshaughn: Nerd.  OK, I took them back out, now merge this before you change your mind :)

PR #173, opened 2017-10-30
Update Metric model for Prometheus

jotak:
- Rename "name" => "displayName"
- Add "family"
- Add "labels"
- Remove unused "type"


jotak:
@jmazzitelli , does it match what you expect?

(PS: I also removed "type", not sure on this one)

jmazzitelli:
This looks OK. As of now, yes, type can be removed since it really serves no purpose. I'm still surprised we have no need for a MetricType object. But since we do not have MetricType, then we do not need a "type" field here. If we ever do add  MetricType (to be attached to ResourceType) then we'll need to add "type" back in to this Metric object. But for now, since we do not have a MetricType object, this looks good.

PR #175, opened 2017-10-30
Can now build -P-ui to optionally omit the ui build. 

jshaughn:
This is primarily to speed up dev builds and as such the ui is still built by default.


jotak:
LGTM

PR #180, opened 2017-11-02
HAWKULAR-1292 Have Inventory update the Prometheus scrape endpoints

lucasponce:
As discussed in HAWKULAR-1292 this PRs filters inventory and when an agent is added, it will create the Prometheus configuration file.
Then, the agent has not to invoke this operation, this happens on registration (or modification) of agents in inventory.
This filter is cheap in terms of performance, so I don't see a problem here.
Also, filters are defined from configuration.
Today there is the assumption that there is a single p8s server, and probably very soon we will play with other topologies, then this scheme can be used to maintain the state which agent is managed  in which p8s server.


lucasponce:
This would need to sync with https://github.com/hawkular/hawkular-agent/pull/399 to expect agent has stored metric endpoint information in its configuration.

jshaughn:
I added a comment in HAWKULAR-1292[](https://issues.jboss.org/browse/HAWKULAR-1292?focusedCommentId=13485085&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13485085)

PR #184, opened 2017-11-03
HAWKULAR-1325 Increase wait time for initialization

lucasponce:
InventoryConfig performs some tasks at init phase that might need more time than standard 5s, so, increasing this timeout will help on slow environments like travis where we have seen this issue.

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryConfig.java
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryConfig.java
@@ -21,8 +21,10 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
 import javax.annotation.PostConstruct;
+import javax.ejb.AccessTimeout;
 import javax.ejb.Singleton;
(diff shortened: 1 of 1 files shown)


jshaughn:
I am seeing issues related to startup, I'll merge this and retest locally.

jmazzitelli:
my only comment is: is this a concern for other ones, not just this getPath? i.e. should we add custom timeouts elsewhere?

lucasponce:
> my only comment is: is this a concern for other ones, not just this getPath? i.e. should we add custom timeouts elsewhere?

It's an initialization, once getPath is resolved, others should be available.

jotak:
I know that it's already merged, but, can we just tell to never timeout, in order to definitely and without doubt eliminate this risk?
If the path is missing, I don't think we have any degraded mode, we just can't start the server.

Or another way would be to separate some of the config - pieces that should be immediately accessible, like the path - from the pieces that need some initialization time.

lucasponce:
Well, I think timeout in this case is not bad.
If it happens is a sympton that there is a problem with the initialization.

jotak:
> It's an initialization, once getPath is resolved, others should be available.
Yes, but if something changes and getPath is not the first methods called from this class anymore, then it would fail again

jotak:
@lucasponce I think the timeout makes sense for the initialization process, but not for getting some simple config like the path.

lucasponce:
ok, I will move the timeout to type level instead of method level.
That will prevent that if the order of injection changes it will hit this issue.

PR #185, opened 2017-11-03
HAWKULAR-1326 Create prometheus config folder before cache init

lucasponce:
- Change target folder

### DIFF
--- a/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryConfig.java
+++ b/hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryConfig.java
@@ -46,6 +46,7 @@ import org.infinispan.query.dsl.QueryFactory;
  */
 @Startup
 @Singleton
+@AccessTimeout(value = 300, unit = TimeUnit.SECONDS)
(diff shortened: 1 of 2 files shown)


lucasponce:
Please @jshaughn, before to merge, let me know if it you can test it and check your scenario.
Thx

hawkular-inventory-parent/hawkular-inventory-service/src/main/java/org/hawkular/inventory/service/InventoryConfig.java
```diff
@@ -75,6 +76,9 @@
     public InventoryConfig() {
         configPath = Paths.get(System.getProperty("jboss.server.config.dir"), "hawkular");
```
jmazzitelli: while we are doing mkdirs in this method, perhaps we should also mkdirs() this hawkular directory just to be safe?

lucasponce: Yes, we can protect here and perform a mkdirs() too.

lucasponce:
@jmazzitelli I addressed your comments, I will merge it and bump services with the SCRDEPS and the change on run-prometheus.sh script.

PR #186, opened 2017-11-06
support our platform metrics again

jmazzitelli:
see https://issues.jboss.org/browse/HAWKULAR-1321
we use this in https://github.com/hawkular/hawkular-agent/pull/406


jshaughn:
I would think we do need these for bare metal, I'm not sure about OS.  I guess if you remove it from the config white-list then they won't be collected.

PR #192, opened 2017-11-16
add expression to Metric API

jmazzitelli:


jmazzitelli:
leaving this for @jshaughn to finish

jotak:
What is it for?

jshaughn:
@jotak There was a long discussion on hawkular IRC, I can send you the log if you like.  But basically the issue is this:  Some metrics do not map to a single time-series.  An example would be "Aggregated Rejected Web Sessions".  This has a family name of "wildfly_deployment_rejected_sessions" and there is a TS for each deployment.  To get a single TS for the metric you you need to sum() all of the TS for the family name (limited by the feed_id).  With hMetrics the old agent actually did the aggregation and reported a single value.  With Prometheus and the jmx-exporter we store the individual TS.  So, we now need to do the aggregation at query time.  When an inventory client asks for this metric's data we need to know the expression to use to get it from Prometheus.  By default it is family_name{labels}.  But in some cases it will be sum(family_name{labels}).  And in others (like per minute) we may need to do some other things.

When querying inventory for data we will perform this as part of the call, the inventory client will just get the data it expects.  In some cases the inventory client may need to query prometheus directly, or include a metric expression in a larger query.  In that case it should be able to get the resolved expression directly from the metric definition.

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Metric.java
```diff
@@ -104,26 +110,35 @@ public static Builder builder() {
     @JsonInclude(Include.NON_NULL)
     private final MetricUnit unit;
 
-    @DocModelProperty(description = "Metric labels.",
+    @DocModelProperty(description = "Metric expression used to evaluate the metric value. If not set, assume $metric.",
```
jshaughn: I think this should always be set, right?

hawkular-inventory-parent/hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/model/JsonTest.java
```diff
@@ -93,6 +94,7 @@ public void deserializeRawResource() throws Exception {
         assertThat(deserialized).isNotNull();
         assertThat(deserialized.getId()).isEqualTo("ID");
         assertThat(deserialized.getMetrics()).extracting(Metric::getDisplayName).containsExactly("metric");
+        assertThat(deserialized.getMetrics()).extracting(Metric::getExpression).containsExactly("sum($metric)");
```
jshaughn: I don't think $metric is a thing, the expression should be resolved.

hawkular-inventory-parent/hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/model/JsonTest.java
```diff
@@ -82,7 +82,8 @@ public void deserializeResultSet() throws Exception {
     public void deserializeRawResource() throws Exception {
         RawResource resource = new RawResource("ID", "FOO", "FEED", "BAR", null,
                 Collections.singletonList(
-                        new Metric("metric", "m.metric", MetricUnit.BYTES, new HashMap<>(), new HashMap<>())),
+                        new Metric("metric", "m.metric", MetricUnit.BYTES, "sum($metric)", new HashMap<>(),
```
jshaughn: And here, no $metric, I think.

hawkular-inventory-parent/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Metric.java
jmazzitelli: if we want to make it required, we can.

hawkular-inventory-parent/hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/model/JsonTest.java
jmazzitelli: yes, now it will be resolved.

jshaughn:
LGTM

jotak:
Thanks @jshaughn for the explanation

PR #194, opened 2017-11-20
fix deploy status pattern - need to whitelist it

jmazzitelli:


jshaughn:
Yes, we will need this.  Although, I think avail mapping is still unhandled in the migration. I will merge when green.

PR #197, opened 2017-11-23
split out domain config from standalone config

jmazzitelli:


jmazzitelli:
This PR is part of a group - these 3 PRs are required for this feature:

https://github.com/hawkular/hawkular-commons/pull/197
https://github.com/hawkular/hawkular-agent/pull/421
https://github.com/hawkular/hawkular-services/pull/244

jshaughn:
This approach looks like it could get tedious, but I'm not sure there is a better alternative given that we can no longer get consolidate metric reporting from the host, and doing it ourselves would be very messy as we'd need to mix dmr and jmx in the agent.  Because this domain stuff is likely only applicable to bare metal, this approach is probably best, and it does distribute the work in a better way.

PR #198, opened 2017-11-23
Enable JDBC events to reach CommandEventListener

tumido:
When I added the JDBC event mapping for MIQ in the related PR, I didn't know JDBC responses are not reaching the CommandEventListener, let's fix it.

I'm not sure if this is the proper fix, please correct me if I'm wrong. I'm trying to test it, but without success so far (I'm not able to build the hawkular-services properly). Can you please help me to test that?

Related to: https://github.com/hawkular/hawkular-services/pull/191
Fixes: [BUG 1515476](https://bugzilla.redhat.com/show_bug.cgi?id=1515476)

cc @jshaughn


jshaughn:
Yes, adding the Interface to the command type should ensure that the CommandResponse is published to JMS and can then be consumed by the listener.  Merging...

PR #200, opened 2017-11-27
HWKCOMMONS-2: Make the Prometheus scrape file location configurable

mwringe:


jmazzitelli:
This looks good to me.

I'll let @lucasponce review and merge since this was code originated by him.

lucasponce:
LGTM
Do we want as well to put outside of the jboss data dir the inventory files ?

mwringe:
@lucasponce I was thinking of just moving the whole jboss data dir to a separate location (eg -Djboss.server.data.dir=/pv). Do you know if its better to move it all across, or selectively just move the inventory pieces there?

lucasponce:
@mwringe both options are good, rpm installations of EAP move/split the data and config from the bin and libraries, probably more. Specifically for hServices inventory and scrape points are located under jboss.server.data.dir, so also it makes sense to have all the data in a same location, I guess. But yes, as scrape points are the integration point between hServices and Prometheus, having a special case can be useful in the future, too.

PR #205, opened 2017-12-12
Hawkular 1346 domain metadata

jmazzitelli:


lucasponce:
LGTM
I wonder we will need it as well in the EAP6 config, isn't it ?

jmazzitelli:
We are not supporting domain mode for EAP6. So there is nothing to do for that version.

lucasponce:
@jmazzitelli last thing, I ran locally the PR and I see some differences:
image
vs
image

lucasponce:
> We are not supporting domain mode for EAP6. So there is nothing to do for that version.

Ok, I didn't remember that. Thanks for the update.

jmazzitelli:
@lucasponce ok, those differences were from the aggregated web server metrics. I added those, too.

lucasponce:
aggregated web server metrics in standalone work fine
```sum(wildfly_deployment_active_sessions{feed_id="attilan"})```

but in the domain wildfly server this metric/expression
```sum(wildfly_deployment_active_sessions{domain_server="server-one",domain_host="master",feed_id="attilan2"})```
doesnt exist, I guess it should exist and returns a 0, like the previous one

jmazzitelli:
> doesnt exist, I guess it should exist and returns a 0, like the previous one

OK, for this one, there is nothing the agent can do. I just looked at how WildFly behaves.

Without ANY deployments, there is NO JMX MBean that WildFly creates that matches the pattern the JMX Exporter needs. These two patterns:

https://github.com/hawkular/hawkular-commons/blob/hawkular-1275/hawkular-inventory-parent/hawkular-inventory-service/src/main/resources/WF10-jmx-exporter.yml#L89-L97

But once you deploy a war or ear to the domain, the slave server WildFly will create an MBean for that deployment  (something like "jboss.as:deployment=my-test.war,subsystem=undertow").

With no deployments, there are no MBeans. The JMX Exporter will not emit any metrics in this case. There is nothing for it to emit. No MBeans == the metrics don't even exist. So Prometheus won't even scrape a "0" value from jmx exporter - there is literally no data.

Thus, for all the aggregate metrics (that aggregate all the deployment MBean data), you will get no data at all because I assume a Prometheus query of "sum()" aggregating over non-existing timeseries results in an empty set of data (not "0").

lucasponce:
> Without ANY deployments, there is NO JMX MBean that WildFly creates that matches the pattern the JMX Exporter needs.

True. Then it seems fine, the empty value will be considered in MiQ as 0 as well.

Ok. Merging.

PR #206, opened 2017-12-14
HAWKULAR-1339 Remove msg message to HawkularTopic

lucasponce:
HawkularTopic was used in BackfillCache feature which is no longer on HAWKULAR-1275, so sending this message on websocket close is not necessary and might cause errors on server log.


jshaughn:
To be honest, I feel like this notification will still be useful somehow in the future, but there is no use case for today, and we can always bring it back. So, LGTM.

